PR #35, opened 2022-08-18
Fix: Change base to remote ref when running test on develop branch

oscar60310:
Circle CI doesn't pull main branch to local, we need to refer to remote branch instead.


kokokuo (approved):
LTGM 👍

PR #37, opened 2022-08-23
Chore: Upload report with flags

oscar60310:
### What's happened

We only run affected packages with CI, so we won't get coverage reports for all packages. But with codecov [merge reports](https://docs.codecov.com/docs/merging-reports) feature, it'll wait for us to send all of reports.

### Solution
Use flags and enable carryforward option, we can use previous reports when some packages are not changed.

https://about.codecov.io/blog/new-product-test-only-what-you-change-with-carryforward-flags/
Flags: https://docs.codecov.com/docs/flags#tracking-flag-coverage-over-time
Carryforward: https://docs.codecov.com/docs/carryforward-flags

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,10 +117,24 @@ jobs:
       - restore_cache:
           keys:
             - codecov-0.1.0_4653
+      - run: rm -rf ./coverage
+        # Clear coverage folder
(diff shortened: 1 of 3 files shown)


kokokuo (approved):
Great feature for CI, LGTM 👍

PR #38, opened 2022-08-23
Feature: Document server - redoc

oscar60310:
### Description
Host [redoc](https://github.com/Redocly/redoc) on `/doc` path (configurable) and provide restful API spec.

### Note
1. For mounting different APIs, I moved Vulcan API under `/api` prefix, e.g. `/user` -> `/api/user` .
2. I [removed](https://github.com/Canner/vulcan/pull/38/files#diff-0bdc30bc0d18f8e9dc4059460c228dc7984a59f3f03d40332457d9a0f1963cdcL36-L40) the fallback data from formaters because they erase the plain text content.


packages/serve/src/lib/middleware/docServerMiddleware.ts
```diff
@@ -0,0 +1,43 @@
...
+import {
+  VulcanInternalExtension,
+  TYPES as CORE_TYPES,
+  DocumentOptions,
+} from '@vulcan-sql/core';
+import { inject, interfaces } from 'inversify';
+import { TYPES } from '@vulcan-sql/serve/containers';
+import * as compose from 'koa-compose';
+
+@VulcanInternalExtension()
```
kokokuo, line 16: How about to setup the all different document server type under one options module name, because seems different document server extensions both use the same `DocumentOptions`.

like below:

```yaml
api-doc:
  spec: [oas3]
  server: [redoc, ...]
```

And your document server middleware read the `doc` options, and according to your `server` to create each document type server ?

packages/serve/src/lib/document-server/redocDocumentServer/redocDocumentServer.ts
```diff
@@ -0,0 +1,86 @@
...
+} from '@vulcan-sql/core';
+import * as nunjucks from 'nunjucks';
+import { Next } from 'koa';
+import * as Router from 'koa-router';
+import { DocumentServer } from '@vulcan-sql/serve/models';
+import * as fs from 'fs';
+import * as path from 'path';
+import { inject } from 'inversify';
+
+@VulcanInternalExtension('redoc')
```
kokokuo, line 18: Follow by above `documentServerMiddleware` suggestion. Here the `@VulcanInternalExtension('redoc')` may could change to `documentServerMiddleware` internal extension module name ?

packages/core/src/models/documentOptions.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export enum DocumentServerType {
+  redoc = 'redoc',
+}
+
+export interface IDocumentOptions {
+  /** Target specification of our APIs, e.g. OpenAPI, Tinyspec ...etc. */
+  specs?: (string | DocumentSpec)[];
+  folderPath?: string;
+  server?: (string | DocumentServerType)[];
```
kokokuo, line 13: Maybe we could rename the propety `server` and `DocumentServerType` e.g: from `server` To the `renderType`, and the `DocumentRenderType` or `DocumentRouterType` `DocumentType`?

Because seems that each document server extension still uses the same port in vulcan server with a different router.

If we use the `server`, it looks like each document server type use a different port to launch.

kokokuo: Similar suggestion for `documentServerMiddleware` and `documentServer` extension, we could change the `server` to another word, e.g: render, router... and so on, to prevent each document server will create a new server with a different port

kokokuo (approved):
Beside some some suggestion, others LGTM 👍

packages/serve/src/lib/middleware/docServerMiddleware.ts
oscar60310: Did you mean to change to property name from `document` to `api-doc` and turn it from internal options to module configuration?

I think it will be great for renaming, but because it a cross-package config, that is, spec for both `build` and `serve` package and server for `serve` package, I'd prefer to use internal options to do some validations.

packages/core/src/models/documentOptions.ts
oscar60310: Thanks, I think router will be better, I'll update all of them.

packages/serve/src/lib/document-server/redocDocumentServer/redocDocumentServer.ts
oscar60310: it makes sense to merge the options of redoc with 'api-doc' options because it is a built-in router, but I'll prefer using internal options instead of module configuration.

oscar60310:
@kokokuo  I've solve the conflicts and issues, thanks a lot.

kokokuo (approved):
LGTM 👍  Thanks ~

PR #42, opened 2022-08-31
Fix: Fix response streams in template

oscar60310:
### What's happened

DataSources returned streams instead of arrays, but we haven't adjusted the template output:
```sql
{% req user %}
select * from users where id = {{ context.params.id }}
{% endreq %}
{{ user.value()[0].name }} --- this will be undefined
```
This PR consume the stream and download data for templates.

### DIFF
--- a/packages/core/src/lib/template-engine/built-in-extensions/query-builder/executorBuilder.ts
+++ b/packages/core/src/lib/template-engine/built-in-extensions/query-builder/executorBuilder.ts
@@ -2,9 +2,36 @@ import {
   FilterBuilder,
   VulcanInternalExtension,
 } from '@vulcan-sql/core/models';
-import { EXECUTE_FILTER_NAME } from './constants';
+import { EXECUTE_COMMAND_NAME, EXECUTE_FILTER_NAME } from './constants';
+import * as nunjucks from 'nunjucks';
+import { ReplaceChildFunc, visitChildren } from '../../extension-utils';
 
 @VulcanInternalExtension()
 export class ExecutorBuilder extends FilterBuilder {
   public filterName = EXECUTE_FILTER_NAME;
+
+  public override onVisit(node: nunjucks.nodes.Node) {
+    visitChildren(node, this.replaceExecuteFunction.bind(this));
+  }
+
+  private replaceExecuteFunction(
+    node: nunjucks.nodes.Node,
+    replace: ReplaceChildFunc
+  ) {
+    if (
+      node instanceof nunjucks.nodes.FunCall &&
+      node.name instanceof nunjucks.nodes.LookupVal &&
+      node.name.val.value === EXECUTE_COMMAND_NAME
+    ) {
+      const args = new nunjucks.nodes.NodeList(node.lineno, node.colno);
+      args.addChild(node.name.target);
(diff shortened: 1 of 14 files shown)


packages/core/src/lib/template-engine/built-in-extensions/query-builder/executorRunner.ts
```diff
@@ -1,19 +1,31 @@
...
-    const builder: IDataQueryBuilder = value;
-    return builder.value();
+    const response = await builder.value();
+
+    // if input value is not a query builder, call the function .value and do nothing.
+    if (!isDataResult(response)) return response;
+
+    const { getData } = response;
+    const dataStream = getData();
+    const data = await streamToArray(dataStream);
```
kokokuo, line 28: If we execute `builder.value()` and consume the stream to the array in the `ExecutorRunner` when play `RuntimeExtension` and it's the final result, then how could we convert the array result to json / csv format in the middleware?

packages/core/src/lib/template-engine/built-in-extensions/query-builder/executorBuilder.ts
```diff
@@ -2,9 +2,36 @@ import {
...
+    ) {
+      const args = new nunjucks.nodes.NodeList(node.lineno, node.colno);
+      args.addChild(node.name.target);
+      const filter = new nunjucks.nodes.Filter(
+        node.lineno,
+        node.colno,
+        new nunjucks.nodes.Symbol(node.lineno, node.colno, EXECUTE_FILTER_NAME),
+        args
+      );
+      replace(filter);
```
kokokuo, line 34: Curious about `replace`, when we call `replace` what do we replace to ?

kokokuo (approved):
Beside some question, others LGTM 👍

packages/core/src/lib/template-engine/built-in-extensions/query-builder/executorBuilder.ts
oscar60310: It is a [callback function](https://github.com/Canner/vulcan/blob/develop/packages/core/src/lib/template-engine/extension-utils/helpers.ts#L29) to remove or replace the current node in order to modify the AST tree, which provided by visitChildren helper.

oscar60310:
Hi @kokokuo , all issues have been resolve, I also found another bug and fix it, https://github.com/Canner/vulcan/pull/42/commits/4a2fdca4d9325feb2e7a510ed6e9cb40ba3acc86

packages/core/src/lib/template-engine/built-in-extensions/query-builder/executorBuilder.ts
kokokuo: Thanks a lot for replying to me

packages/core/src/lib/template-engine/built-in-extensions/query-builder/executorRunner.ts
kokokuo: Okay, thanks

kokokuo (approved):
LGTM 👍  cc @oscar60310

PR #47, opened 2022-09-05
Feature: provide getting auth token and user profile api for client-server used.

kokokuo:
## Description
Provide an getting token info POST API `/auth/token` and get user profile GET API `/auth/user-profile` for client-server to authenticate and get user information. Otherwise, we also support providing auth credentials data ( e.g: token ) from query string or payload. So it not only has one way to provide by `Authorization` in the header.

We will talk about it in two sections.

##  How to use `/auth/token` and `/auth/user-profile` API
The `/auth/token` POST API is used to get tokens by passing user identity information, e.g: `username` or `password`. After you get the token, you could continue to get the current user's basic information by GET `/auth/user-profile` with the above got token.

Also if you would like to query the data endpoint API, you also need the token.

### Set `auth` options to make API work first.

The `type` value should provide satisfy the key name under `auth` options, e.g: `basic`, `simple-token`, `password-file`.. and so on. 

#### Sample 1

```yaml
auth:
 options:
  basic:
   htpasswd-file: 
     ....
```

Then you could access auth and get a token with API `/auth/token` endpoint. 
```JSON
{
  "type": "basic"
  // below is the auth type needed for other payload data, e.g: username, password
}
```

#### Sample 2
If you also added `password-file` auth type options in settings:
```yaml
auth:
 options:
  basic:
   htpasswd-file: 
    path: './yyyy'
  simple-token:
   - name: user1
     token: '....'
```

Then the auth API `/auth/token` with `type` could use `simple-token` and `basic`.

### Error message
If you do not provide `type` query string, it will show `400` status code with the message, like below:
```json
{
  "message": "auth type 'xxx',in options not supported, authenticator only supported"
}
```

### Other payloads like username, password
Current built-in authenticators: `simple-token`, `password-file`  and `basic` both need to be given `username` and `password` fields in the payload.

```json
{
 "type": "basic"
 "username": "user1",
 "passowrd": "test1"
}
```

 If the client missed some field e.g: `username` or `password`  it will throw an error message with status `400`:

```json
{
  "message": "error message....."
}
```

After you get the token e.g: `dXNlcjE6dGVzdDE=`, then you could get the user profile following the above `/auth/user-profile` GET API we talked. 

Just call the GET API `/auth/user-profile` and set the `Authorization` with token value e.g: `dXNlcjE6dGVzdDE=` in request header.

### Customize your Authenticator
If the above built-in `basic`,  `password-files`, `simple-token` authenticator could not satisfy, we could define the `authenticator`, you should inherit `BaseAuthenticator` and implement the `getTokenInfo` and `authCredential` method.

## Support providing token or other related auth credentials data from query string or payload

We also define another middleware called `authSourceNormalizerMiddleware` to support clients who could pass the auth credentials data query string or payload.

### Set the `auth-source` options

Default, the client could through `auth` key and base64 with JSON stringify flow to encode your `Authroizartion` key with value. like below:

```bash
/auth/user-profile?auth=Base64(JSON.stringfiy({'Authorization': `Basic dXNlcjE6dGVzdDE=`})) 
```

Then we will find the `auth` in the query string and check if the format is Base64 or not, if yes, we will try to decode it and move it to the header, but, currently, we only accept the `Authorization` key name in the object. ( We will provide a whitelist feature to make user could define the acceptable key name in the object )

If you would like to change the key name like `x-auth` to replace the `auth` key, or you would like to set `x-auth` by payload, the  you could define it in `auth-source` options:

```yaml
auth-source:
  key:'x-auth'  # default is "auth" key name
  in: 'query'    # default is "query",  and we support "payload" or "query" two types
```

like above, then when you send a query to the data endpoint e.g: `/orders`, it looks like below:

```bash
/orders?x-auth=Base64(JSON.stringfiy({'Authorization': `Basic dXNlcjE6dGVzdDE=`})) 
```

## Commit Message
* 04eb9f4 feat(serve): provide auth identity route for client server used.
* f13dc42 fix(serve): add error handler for an auth route middleware
* 016656a fix(serve): move `server.close` to after each function in test cases
* 1a0930a fix(serve): move the auth credential middleware checking option logistic to `onActivate` method
* 044b649 fix(serve): add checking enable condition and option value in `onActivate` method of auth route middleware
* 7c3f9b4 fix(serve): the /auth route still need to contain `Authorization` header issue
* 4dfd5cc fix(serve): remove checking user credential logistic and only generate token
* 7f6d397 fix(serve): change get method of `/auth/token` endpoint to post method
* 14f3bb9 feat(serve): add `/auth/user-profile` api to get user profile.
* ac2d8d5 feat(serve): add auth source normalizer middleware to support query string source auth data
* 258157f fix(serve): import 'koa-bodyparser' in the each authenticator to make jest run correct.
* 2bce5e8 fix(serve,labs): refactor error message when not provide auth types in options, and set disable for auth in `vulcan.yaml` template in labs project


packages/serve/src/lib/middleware/auth/authRouteMiddleware.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  public async handle(context: KoaContext, next: Next) {
+    // return to stop the middleware, if disabled
+    if (!this.enabled) return next();
+
+    if (isEmpty(this.options))
+      throw new Error(
+        'please set at least one auth type and user credential when you enable the "auth" options.'
+      );
```
oscar60310, line 69: Can we check the option at `onActivate` function too?
```ts
if (this.enable && isEmpty(xxxxx
```

packages/serve/src/lib/middleware/auth/authRouteMiddleware.ts
```diff
@@ -0,0 +1,112 @@
...
+          `The auth type "${type}" in options not supported, authenticator only supported ${authIds}.`
+        );
+    });
+
+    for (const authId of authIds) {
+      const authenticator = this.authenticators[authId];
+      if (authenticator.activate) await authenticator.activate();
+    }
+    // setup route
+    this.setAuthRoute();
```
oscar60310, line 59: Can we only mount our routes only when extension is enabled?
```ts
if(this.enable) this.setAuthRoute();
```

packages/serve/src/lib/middleware/auth/authRouteMiddleware.ts
```diff
@@ -0,0 +1,112 @@
...
+      throw new Error(
+        'please set at least one auth type and user credential when you enable the "auth" options.'
+      );
+
+    // run each auth route
+    await this.router.routes()(context, next);
+  }
+
+  private setAuthRoute() {
+    this.router.get(`/auth`, async (context: KoaContext, next) => {
```
oscar60310, line 76: I'd prefer using POST instead of GET to send our credentials.

oscar60310: Ought this route to be public? For now, it's protected by `authCredentialMiddleware`, that is, we have to send credentials before knowing how to get one.

❌
curl --location --request GET 'http://localhost:3000/auth?type=basic&username=ivan&password=123'

✅
curl --location --request GET 'http://localhost:3000/auth?type=basic&username=ivan&password=123' \
--header 'Authorization: Basic aXZhbjoxMjM='

packages/serve/test/middlewares/built-in-middlewares/authRouteMiddleware.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+    const server = await runServer(options, stubAuthenticators);
+
+    // Act
+    const request = supertest(server).get(`/auth`);
+    const response = await request;
+
+    // Assert
+    expect(response.statusCode).toEqual(400);
+    expect(response.body).toEqual(expected);
+    server.close();
```
oscar60310, line 163: We should close server in afterEach hooks, or they become dangling if test cases failed.

oscar60310 (approved):
Other part LGTM.

packages/serve/src/lib/middleware/auth/authRouteMiddleware.ts
kokokuo: Thanks for @oscar60310 suggesting, it has been fixed.

packages/serve/src/lib/middleware/auth/authRouteMiddleware.ts
kokokuo: Thanks for @oscar60310 suggesting. it has been fixed.

packages/serve/src/lib/middleware/auth/authCredentialMiddleware.ts
```diff
@@ -54,22 +50,25 @@ export class AuthMiddleware extends BuiltInMiddleware<AuthOptions> {
     if (!this.enabled) return next();
 
     const options = (this.getOptions() as AuthOptions) || {};
-    if (isEmpty(options)) return next();
+    if (isEmpty(options))
+      throw new Error(
+        'please set at least one auth type and user credential when you enable the "auth" options.'
+      );
```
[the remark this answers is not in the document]
kokokuo: Thanks for @oscar60310 suggesting. it has been fixed.

packages/serve/src/lib/middleware/auth/authRouteMiddleware.ts
kokokuo: Thanks for @oscar60310 finding the issue. it has been fixed.

packages/serve/test/middlewares/built-in-middlewares/authRouteMiddleware.spec.ts
kokokuo: Thanks for @oscar60310 suggestion. it has been fixed.

packages/serve/src/lib/auth/httpBasicAuthenticator.ts
```diff
@@ -87,26 +87,45 @@ export class BasicAuthenticator extends BaseAuthenticator<BasicOptions> {
...
+      !(md5(password) === this.usersCredentials[username].md5Password)
+    )
+      throw new Error('authenticate user identity failed.');
+
+    const token = Buffer.from(`${username}:${password}`).toString('base64');
+
+    return {
+      token: token,
+    };
+  }
```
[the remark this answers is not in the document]
kokokuo: Thanks @oscar60310 the for suggestion, after discussion I have made it only generate the token, and provide another API `/auth/user-profile` to get the user data with the token which got from `/auth/token`

packages/serve/src/lib/middleware/auth/authRouteMiddleware.ts
kokokuo: I also add more test cases related to authentication in the `integration-testing` package and change the structure.

kokokuo:
Hi @oscar60310, it has been fixed and provide new API for getting user profile, thanks for discussing with me and giving suggestion 😃

oscar60310 (approved):

PR #48, opened 2022-09-06
Feature: CLI with docker

oscar60310:
Closing #49 

## Usage
1. Create a new folder and cd into it.
    ```bash
    mkdir vulcan-project && cd vulcan-project
    ```
2. Set alias (optional)
    ```bash
    alias vulcan="docker run -it --rm -v ${PWD}:/usr/app -p 3000:3000 ghcr.io/canner/vulcan/cli:dev"
    ```
3. Initialize project
    ```bash
    vulcan init .
    ```
5. Start server
    ```bash
    vulcan start
    ```

## Notes
I also add a script for bumping dev, beta, and latest versions. For incoming feature #53

### DIFF
--- a/tools/scripts/publishDocker.mjs
+++ b/tools/scripts/publishDocker.mjs
@@ -3,6 +3,8 @@ import path from 'path';
 import fs from 'fs';
 import { execSync } from 'child_process';
 
+// node publishDocker.mjs <tag> <version>
+
 if (process.env.READY_FOR_PUBLISH !== 'true') {
   console.log(`Set env READY_FOR_PUBLISH=true before running publish commands.`)
   process.exit(1);
--- a/tools/scripts/publishDocker.mjs
+++ b/tools/scripts/publishDocker.mjs
@@ -1,0 +1,22 @@
+import { getReleaseTag, getVersionByArguments } from './version.mjs';
+import path from 'path';
+import fs from 'fs';
+import { execSync } from 'child_process';
+
+if (process.env.READY_FOR_PUBLISH !== 'true') {
+  console.log(`Set env READY_FOR_PUBLISH=true before running publish commands.`)
+  process.exit(1);
+}
+
+const packageName = JSON.parse(fs.readFileSync(path.resolve(process.cwd(), 'package.json'), 'utf-8')).name.replace('@vulcan-sql/', '');
+if (!packageName) throw new Error(`Can't find package name`);
+const version = getVersionByArguments();
+const tag = getReleaseTag();
+
+// run docker build xxx -t vulcan-sql/<package-name> before publish
+// we tag it with url and version
+execSync(`docker tag vulcan-sql/${packageName} ghcr.io/canner/vulcan-sql/${packageName}:${version}`);
+execSync(`docker tag vulcan-sql/${packageName} ghcr.io/canner/vulcan-sql/${packageName}:${tag}`);
+// push
+execSync(`docker push ghcr.io/canner/vulcan-sql/${packageName}:${version}`);
+execSync(`docker push ghcr.io/canner/vulcan-sql/${packageName}:${tag}`);
--- a/tools/scripts/publishDocker.mjs
+++ b/tools/scripts/publishDocker.mjs
@@ -13,10 +13,10 @@ if (!packageName) throw new Error(`Can't find package name`);
 const version = getVersionByArguments();
 const tag = getReleaseTag();
 
-// run docker build xxx -t vulcan-sql/<package-name> before publish
+// run docker build xxx -t vulcan-sql/<package-name>:<version> before publish
 // we tag it with url and version
-execSync(`docker tag vulcan-sql/${packageName} ghcr.io/canner/vulcan-sql/${packageName}:${version}`);
-execSync(`docker tag vulcan-sql/${packageName} ghcr.io/canner/vulcan-sql/${packageName}:${tag}`);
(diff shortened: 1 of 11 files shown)


tools/scripts/publishDocker.mjs
```diff
@@ -0,0 +1,22 @@
...
+const version = getVersionByArguments();
+const tag = getReleaseTag();
+
+// run docker build xxx -t vulcan-sql/<package-name> before publish
+// we tag it with url and version
+execSync(`docker tag vulcan-sql/${packageName} ghcr.io/canner/vulcan/${packageName}:${version}`);
+execSync(`docker tag vulcan-sql/${packageName} ghcr.io/canner/vulcan/${packageName}:${tag}`);
+// push
+execSync(`docker push ghcr.io/canner/vulcan/${packageName}:${version}`);
+execSync(`docker push ghcr.io/canner/vulcan/${packageName}:${tag}`);
```
kokokuo, line 22: Why do you make the docker image have different tag by `version` and `tag` ? Is it any case that could be needed? 

Because only follow the code, since the `getReleaseTag` function, looks like it will not related to `version` in `package.json`, but we allow user could tag one of the `dev`, `beta`, `lastest`, so usually user will also add the version before or after the tag e.g: **dev-0.1.0**, **0.1.0-dev**, but when the `version` in the `package.json` is `0.0.9` it will make it a little confusing.

kokokuo (commented):
Beside some question, Others LGTM 👍 

Btw, maybe we could write a README to introduce what tag we allow users to type and what scenario they could use it.

tools/scripts/publishDocker.mjs
oscar60310: > Why do you make the docker image have different tag by `version` and `tag` ? Is it any case that could be needed?

I'd like to move `dev`, `beta`, and `latest` tags to the latest image we pushed, for example:
Today
0.1.0-dev.20220907.0 dev
Tomorrow
0.1.0-dev.20220907.0
0.1.0-dev.20220908.0 dev


> Because only follow the code, since the `getReleaseTag` function, looks like it will not related to `version` in `package.json`, but we allow user could tag one of the `dev`, `beta`, `lastest`, so usually user will also add the version before or after the tag e.g: **dev-0.1.0**, **0.1.0-dev**, but when the `version` in the `package.json` is `0.0.9` it will make it a little confusing.

I should update package.json with the desired version before building, I'll fix it.

oscar60310:
Hi @kokokuo , I've updated the scripts, modify package before building.
We can check it by:
 docker run --rm ghcr.io/canner/vulcan-sql/cli:dev version

oscar60310:
> Btw, maybe we could write a README to introduce what tag we allow users to type and what scenario they could use it.

These three tags follow our release channels design, only few users will use publish command, and user should use `nx` commands instead of calling these script files directly because of the dependencies and PWD. So maybe we can describe these commands along with our release channels instead of an individual document.

kokokuo:
> > Btw, maybe we could write a README to introduce what tag we allow users to type and what scenario they could use it.
> 
> These three tags follow our release channels design, only few users will use publish command, and user should use `nx` commands instead of calling these script files directly because of the dependencies and PWD. So maybe we can describe these commands along with our release channels instead of an individual document.

I'm not sure where is the release channels you said, our [github repo release channels](https://github.com/Canner/vulcan-sql/releases) or a release channel in our VulcanSQL discord?
Btw, you also said **PWD**, what is the PWD abbreviation here you would like to talk? Linux command PWD ? or other meaning?

kokokuo:
> docker run --rm ghcr.io/canner/vulcan-sql/cli:dev version

Should we don't restrict our release tag version must larger than versions of `package.json` and not restrict the format of tag ? so if other developers type different words/numbers behind `dev`, `beta`, `latest ` also work? e.g:  add codename like ubuntu , `latest-candybar`

oscar60310:
> > > Btw, maybe we could write a README to introduce what tag we allow users to type and what scenario they could use it.
> > 
> > 
> > These three tags follow our release channels design, only few users will use publish command, and user should use `nx` commands instead of calling these script files directly because of the dependencies and PWD. So maybe we can describe these commands along with our release channels instead of an individual document.
> 
> I'm not sure where is the release channels you said, our [github repo release channels](https://github.com/Canner/vulcan-sql/releases) or a release channel in our VulcanSQL discord? Btw, you also said **PWD**, what is the PWD abbreviation here you would like to talk? Linux command PWD ? or other meaning?

For release channels I simply meant the release tags we used for NPM: 
- `latest` for stable releases.
- `beta` for testing releases.
- `dev` for nightly builds.

PWD stands for pathname of the current working directory, I've set different working directory for each commands, it'll be difficult to call the scripts file directly without the help from `project.json`.
https://github.com/Canner/vulcan-sql/blob/develop/packages/build/project.json#L53

oscar60310:
> > docker run --rm ghcr.io/canner/vulcan-sql/cli:dev version
> 
> Should we don't restrict our release tag version must larger than versions of `package.json` and not restrict the format of tag ? so if other developers type different words/numbers behind `dev`, `beta`, `latest ` also work? e.g: add codename like ubuntu , `latest-candybar`

I'll add another flag `--version` to let users to specify the desired version instead of auto generating.
For example: release core package with `dev` tag and version `1.2.3`
```bash
nx publish:docker core --tag dev --version 1.2.3
```

oscar60310:
Hi @kokokuo  I've updated our publish/build scripts, please have a look at the new commit, thanks for the suggestion! I'll update the NPM publish script too at #68 .

kokokuo (approved):
LGTM 💯

PR #75, opened 2022-09-15
Feature: Data source profiles

### LINKED ISSUE #71: Data source profiles
We need to load data source connection information to execute queries, e.g. host, port, dbname, dbuser, dbpassword ...  for Postgres.

1. Users can set `profiles` property in their project config file (vulcan.yaml), it contains all the profiles' paths, in order of priority, from top to bottom.
    ```yaml
    profiles:
      - some/path/to/profiles.yaml
      - another/path/to/profiles.yaml
    ```

2. Two default paths will append at the end automatically: `profiles.yaml` and `~/.vulcan/profiles.yaml`. For example:
    ```yaml
    profiles:
      - some/path/to/profiles.yaml
      - another/path/to/profiles.yaml
    ```
    We'll try to find profiles from `some/path/to/profiles.yaml` first, then `another/path/to/profiles.yaml`, `profiles.yaml`, and  `~/.vulcan/profiles.yaml`

3. Profiles are duplicated by their names.
4. This is an example of a profile file:
    ```yaml
    - name: pg-admin # required. the name of this profile
       type: pg # required. which driver we should use
       connection: # required. connection info, which depends on drivers
          host: example.com
          username: vulcan
          password: xxxx
    - name: duck
      type: duck
      connection:
          path: ./path/to/db
    ```
5. We'll need to set which profile we want to use at endpoints' schemas:
    ```yaml
    urlPath: /users
    profile: pg-admin
    ```

## **Additional context**
1. Switching profiles by environment (production, dev ...) might be implemented in the future.
2. We'll implement authorization feature soon, we'll set multiple profiles for one endpoint at that time.

oscar60310:
## Description
I saved the profiles in data sources, and the relationships among components are:
```
Server/Builder --- Template engine --- Executor --- Builder --- Data Source --- Profiles
```

We need to indicate our profile lookup information at project config (vulcan.yaml):
```yaml
profiles:
   - /path/to/profiles.yaml
   - type: LocalFile
     options:
        path: /path/to/profiles.yaml
   - type: other reader # We can create other readers ... e.g. read from vault
     options:
        foo: bar
```

In the profile file (profiles.yaml), we need to list all the profiles and their connection info:
```yaml
- name: duck
  type: duckdb
  connection:
    persistent-path: ./test-data/moma.db
    log-queries: true
    log-parameters: true
```

And the last, we need to tell Vulcan which profile you want to use in schema:
```yaml
urlPath: /artist/:id/works
request:
  - fieldName: id
    fieldIn: path
    description: constituent id
    validators:
      - required
sample:
  parameters:
    id: '1'
  profile: duck
profile: duck

```
## Issue ticket number

closes #71 

## Additional Context

I created a metadata helper (NunjucksExecutionMetadata) to carry global context instead of read it from the context object directly:
Before
```ts
// Set value
templateEngine.execute(xxx. {[SOME_GLOBAL_KEY]: 'some-value'})
// Use value
public async run({ context, args, contentArgs,}: TagRunnerOptions) {
    context.lookup(SOME_GLOBAL_KEY);
}

```
After
```ts
// Set value
const metadata = new NunjucksExecutionMetadata({profile: 'some-value'});
templateEngine.execute(xxx. metadata.dump())
// Use value
public async run({ context, args, contentArgs, metadata}: TagRunnerOptions) {
    metadata.getProfileName();
}

```


kokokuo (approved):
LGTM 👍 👍 👍

PR #76, opened 2022-09-16
Feature: Authorization

### LINKED ISSUE #72: Authorization
After #71, we have profiles for data sources, we want to do access control for them. For example, admin users can access `pg-admin` profile, but non-admin users can only access `pg-no-admin` profile.

We can implement a sample attribute-based access control (ABAC) for our profiles. 
1. For each profile, we can set the `allow` property to indicate what attributes users should have.
    For example, only the user with name `admin` can access this profile:
    ```yaml
    - name: 'pg-admin'
      driver: 'pg'
      connection: xx
      allow:
        - name: admin
    ```
2. Attribute filters are also supported. For example, user should have an attribute `group` and its value should be `admin`:
    ```yaml
    - name: 'pg-admin'
      driver: 'pg'
      connection: xx
      allow:
        - attribute:
              name: group
              value: admin
    ````
3. Wildcards are supported.
    ```yaml
    - name: 'pg-admin'
      driver: 'pg'
      connection: xx
      allow:
        - name: admin-*
    ```
4. Single allow condition are combined with AND logic
   Only the user who has name `admin` and has `group` attribute with value `admin` can access this profile.
     ```yaml
    - name: 'pg-admin'
      driver: 'pg'
      connection: xx
      allow:
        - name: admin
          attribute:
              name: group
              value: admin 
    ```
   Multiple allow conditions are combined with OR logic.
   `admin`, `someoneelse`, and those who have `group` attribute with value `admin` can access this profile.
     ```yaml
    - name: 'pg-admin'
      driver: 'pg'
      connection: xx
      allow:
        - name: admin
        - name: someoneelse
        - attribute:
              name: group
              value: admin 
    ```
5. We can specify multiple profiles on a single schema now, from top to bottom, users use the first qualified profile. If users can't use any of them, `403` error should be thrown.
## **Additional context**

oscar60310:
## Description
I've implemented the logic of #72 , we only need to inject `evaluator` and provide the user information:
```ts
   const profile = this.evaluator.evaluateProfile(user, profiles);
    if (!profile)
      // Should be 403
      throw new Error(`Forbidden`);
```
I also pass the profile name to template context, we can use `{{ context.profileName }}` to get it.
## Issue ticket number

closes #72

### DIFF
--- a/labs/playground1/profile.yaml
+++ b/labs/playground1/profile.yaml
@@ -4,3 +4,4 @@
     persistent-path: ./test-data/moma.db
     log-queries: true
     log-parameters: true
+  allow: '*'
--- a/packages/build/src/lib/schema-parser/middleware/checkProfile.ts
+++ b/packages/build/src/lib/schema-parser/middleware/checkProfile.ts
@@ -14,19 +14,25 @@ export class CheckProfile extends SchemaParserMiddleware {
   }
 
   public async handle(schemas: RawAPISchema, next: () => Promise<void>) {
+    if (!schemas.profiles && schemas.profile) {
+      schemas.profiles = [schemas.profile];
+    }
+
     await next();
     const transformedSchemas = schemas as APISchema;
-    if (!transformedSchemas.profile)
+    if (!transformedSchemas.profiles)
       throw new Error(
         `The profile of schema ${transformedSchemas.urlPath} is not defined`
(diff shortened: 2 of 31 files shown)


kokokuo (approved):
LGTM 👍 👍 👍

PR #78, opened 2022-09-21
Feature: New driver - pg

### LINKED ISSUE #56: New driver: Postgres - statement
Postgres is one of the most commonly used warehouses, we should add driver support.

Provide Postgres driver as an individual extension.

## **Additional context**
We only support executing queries and preparing statements now, so the driver can ignore the builder commands (`ExecuteOptions.operations`)

oscar60310:
## Description

- Implement data source for Postgres.
- I changed the default data source to "mock" because it was "pg".
- Destroy the source stream when Koa destroyed the downstream, it's important for pg driver because it needs to release the connection to the pool.

## Issue ticket number

closes #56 

## Additional Context
- I used a real pg server in docker to run the test, so our unit tests need Docker env to run.
- I changed the CI environment from docker to VM in order to run docker services


oscar60310:
@kokokuo  I've solved the conflicts, and this PR is ready for review.

kokokuo (approved):
LGTM 👍 👍

kokokuo:
@oscar60310, just have a question after I also develop another data source, why do we need to do the `typeMapper` and convert the PG original type to typescript matched type?

oscar60310:
> @oscar60310, just have a question after I also develop another data source, why do we need to do the `typeMapper` and convert the postgres original type and snowflake to typescript matched type?

For generating the response schema in the specification.
https://github.com/Canner/vulcan-sql/blob/develop/packages/build/src/lib/schema-parser/middleware/responseSampler.ts#L48-L64

PG (or other data sources) has lots of data types, e.g. int2, int4, etc., but our API spec only has some basic types: `string`, `number`, etc. So the connectors have to tell Vulcan core how should we map the corresponding types, that is, make the types simpler to understand.

kokokuo:
> > @oscar60310, just have a question after I also develop another data source, why do we need to do the `typeMapper` and convert the postgres original type and snowflake to typescript matched type?
> 
> 
> 
> For generating the response schema in the specification.
> 
> https://github.com/Canner/vulcan-sql/blob/develop/packages/build/src/lib/schema-parser/middleware/responseSampler.ts#L48-L64
> 
> 
> 
> PG (or other data sources) has lots of data types, e.g. int2, int4, etc., but our API spec only has some basic types: `string`, `number`, etc. So the connectors have to tell Vulcan core how should we map the corresponding types, that is, make the types simpler to understand.
> 
> 
> 
> 

Thanks so much!!!! I'm working ClickHouse now, it really helpful for me to handle it😀

PR #81, opened 2022-09-22
Fix: Fix dependencies issues

oscar60310:
## Description
The package.json of serve package missed two packages:
- redoc: because we imported it at run time only. I won't import it because we only use the bundle result, but not the source.
- bcryptjs: because we set it to dev dependencies

This PR fix these issues.

## Issue ticket number
N/A quick fix

## Additional Context
Related to #38 #26

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "private": true,
   "dependencies": {
     "@koa/cors": "^3.3.0",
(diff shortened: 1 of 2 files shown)


kokokuo (approved):
LGTM 👍

PR #82, opened 2022-09-22
Feature: production build

### LINKED ISSUE #50: Production build with Node and Docker
We need a production build that only includes necessary context, e.g. artifacts, production dependency, serve package ...etc.

Provide `package` command to do production builds, and use `--output` flag to indicate the output type. We can implement node and docker output first:

- Node:
  Build: `vulcan package --output node`
  Execute: `node ./dist/index.js`
- Docker
  Build: `vulcan package --output node && docker build ./dist`
  Execute `docker run <image`

## **Additional context**
- We should prepare the resources e.g. dockerfile and all the contents then print the "docker build" commands to users instead of calling docker API directly. Because we might not able to call docker API from Vulcan CLI, e.g. using CLI in docker #49

oscar60310:
## Description
Provide `package` command to do production builds, and use `--output` flag to indicate the output type. 

- Node:
  Build: `vulcan package --output node`
  Execute: `npm install && node ./dist/index.js`
- Docker
  Build: `vulcan package --output node && docker build ./dist`
  Execute `docker run xxx`

## Issue ticket number

closes #50 

## Additional Context
- I update our artifact builder.  We need to use `getArtifact` to get the content now. cc @wwwy3y3 
  ```ts
    await artifactBuilder.load();
    const templates = artifactBuilder.getArtifact(
      BuiltInArtifactKeys.Templates
    );
    const schemas = artifactBuilder.getArtifact<APISchema[]>(
      BuiltInArtifactKeys.Schemas
    );
  ```
- Breaking changes:
   - Remove `spec-xxx.yaml` files, they'll be wrapped into result.json now.


kokokuo (approved):
LGTM 👍 👍

PR #87, opened 2022-09-23
Fix: Fix format error when url is not found

### LINKED ISSUE #84: Fix format error when url is not found
## Environment
dev

## What's wrong?
Vulcan server throws format errors when the URL is not found.

```
Error: Url ending format not matched in "formats" options
```

It should do nothing with URLs that are not found.

## Reproducing steps
1. Start server.
2. curl localhost:3000/some/uirl

oscar60310:
## Description
Format middleware now only handles Vulcan APIs.

## Issue ticket number

closes #84

## Additional Context
- Hardcoding API in middleware is a temporary solution, we should use a configurable path instead.


kokokuo (approved):
LGTM 👍 👍  Thanks so much for finding and fixing the formatter issues about formatting all apis !

PR #88, opened 2022-09-23
Fix: Parse context.params instead of params for parameter metadata

### LINKED ISSUE #85: Parse context.params instead of params for parameter metadata
## Environment
dev

## What's wrong?
When we add parameters in templates but not in schemas, the document doesn't get these parameters.

It should get these parameters with default validators.

## Reproducing steps
1. Add some parameters like `{{ context.params.id }}` but not in schema definitions.
2. Build Vulcan and check the document.

oscar60310:
## Description
- lookup context.params.x instead of params.x
- add parameters to schema instead of throw error when they are not found

## Issue ticket number

closes #85


kokokuo (approved):
LGTM 👍 👍 👍

PR #86, opened 2022-09-23
Fix: Update CLI initialized files to align the new design

oscar60310:
## Description
Update these SQL files, schemas, and config.

- move build package to dev dependencies
- update sample queries
- add empty profiles.yaml

## Issue ticket number

closes #83

PR #89, opened 2022-09-24
Feature: Throw errors with proper http code

### LINKED ISSUE #61: Server error handling for restful APIs
We haven't had a proper way to handle errors for restful API, that is, we'll always throw "500 internal server error" for any errors.

Catch these errors and set the proper HTTP status code. We can wrap our errors when they occurred and set the desired HTTP code there.

## **Additional context**

oscar60310:
## Description
1. Custom errors
    I've created 4 errors for us:
    ```ts
    /** Expected errors, which is caused by users */
    export class UserError extends VulcanError {
      constructor(message?: string, options?: VulcanErrorOptions) {
        super(message, {
          ...options,
          httpCode: options?.httpCode || 400,
          code: options?.code || 'vulcan.userError',
          exitCode: options?.exitCode || 2,
        });
      }
    }
    
    /** Unexpected errors, which is caused by internal issues */
    export class InternalError extends VulcanError {
      constructor(message?: string, options?: VulcanErrorOptions) {
        super(message, {
          ...options,
          httpCode: options?.httpCode || 500,
          code: options?.code || 'vulcan.internalError',
          exitCode: options?.exitCode || 3,
        });
      }
    }
    
    /** The configurations e.g. vulcan.yaml, user.yaml ... are incorrect */
    export class ConfigurationError extends InternalError {
      constructor(message?: string, options?: VulcanErrorOptions) {
        super(message, {
          ...options,
          code: options?.code || 'vulcan.configError',
        });
      }
    }
    
    /** Error from template syntax */
    export class TemplateError extends InternalError {
      constructor(message?: string, options?: VulcanErrorOptions) {
        super(message, {
          ...options,
          code: options?.code || 'vulcan.templateError',
        });
      }
    }
    ```
2. Catch errors and set proper HTTP code and body.
3. Corrected the order of middlewares.

## Issue ticket number

closes #61

### DIFF
--- a/packages/build/src/containers/container.ts
+++ b/packages/build/src/containers/container.ts
@@ -1,5 +1,5 @@
 import { Container as InversifyContainer } from 'inversify';
-import { Container as CoreContainer } from '@vulcan-sql/core';
+import { Container as CoreContainer, InternalError } from '@vulcan-sql/core';
 import { IBuildOptions } from '@vulcan-sql/build/models';
 import {
   documentGeneratorModule,
@@ -15,7 +15,7 @@ export class Container {
   public get<T>(type: symbol) {
     const instance = this.inversifyContainer?.get<T>(type);
     if (!instance)
-      throw new Error(`Cannot resolve ${type.toString()} in container`);
+      throw new InternalError(`Cannot resolve ${type.toString()} in container`);
     return instance;
   }
 
--- a/packages/build/src/lib/document-generator/spec-generator/oas3/oas3SpecGenerator.ts
+++ b/packages/build/src/lib/document-generator/spec-generator/oas3/oas3SpecGenerator.ts
@@ -17,6 +17,7 @@ import {
   VulcanExtensionId,
   VulcanInternalExtension,
   DocumentSpec,
+  ConfigurationError,
 } from '@vulcan-sql/core';
 import { isEmpty } from 'lodash';
 
@@ -104,7 +105,9 @@ export class OAS3SpecGenerator extends SpecGenerator<oas3.OpenAPIObject> {
       case FieldInType.QUERY:
         return 'query';
       default:
-        throw new Error(`FieldInType ${fieldInType} is not supported`);
+        throw new ConfigurationError(
+          `FieldInType ${fieldInType} is not supported`
+        );
     }
   }
 
@@ -145,7 +148,9 @@ export class OAS3SpecGenerator extends SpecGenerator<oas3.OpenAPIObject> {
       case FieldDataType.STRING:
         return 'string';
       default:
-        throw new Error(`FieldDataType ${fieldDataType} is not supported`);
+        throw new ConfigurationError(
+          `FieldDataType ${fieldDataType} is not supported`
+        );
     }
   }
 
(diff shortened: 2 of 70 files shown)


kokokuo:
I discover one thing and would like to check does it was missed, The CLI init files `user.yaml` seems not to have `sample` in #86 [code href](https://github.com/Canner/vulcan-sql/pull/86/files#diff-dae705d89fb59e4513ec3d96f19f29bc5c20c67059fca22cb67d8307bd25b44bR9 )
Should it also need to add?

oscar60310:
> I discover one thing and would like to check does it was missed, The CLI init files `user.yaml` seems not to have `sample` in #86 [code href](https://github.com/Canner/vulcan-sql/pull/86/files#diff-dae705d89fb59e4513ec3d96f19f29bc5c20c67059fca22cb67d8307bd25b44bR9) . Should it also need to add `sample` ?

Thanks for pointing out, that property is optional, builder won't send sample queries without setting it. So I think it's ok to leave it blank in init templates.

kokokuo:
> > I discover one thing and would like to check does it was missed, The CLI init files `user.yaml` seems not to have `sample` in #86 [code href](https://github.com/Canner/vulcan-sql/pull/86/files#diff-dae705d89fb59e4513ec3d96f19f29bc5c20c67059fca22cb67d8307bd25b44bR9) . Should it also need to add `sample` ?
> 
> 
> 
> Thanks for pointing out, that property is optional, builder won't send sample queries without setting it. So I think it's ok to leave it blank in init templates.
> 
> 
> 
> 

If not set it, does the api spec still generated but not appear the response info in the api document?

oscar60310:
> > > I discover one thing and would like to check does it was missed, The CLI init files `user.yaml` seems not to have `sample` in #86 [code href](https://github.com/Canner/vulcan-sql/pull/86/files#diff-dae705d89fb59e4513ec3d96f19f29bc5c20c67059fca22cb67d8307bd25b44bR9) . Should it also need to add `sample` ?
> > 
> > 
> > Thanks for pointing out, that property is optional, builder won't send sample queries without setting it. So I think it's ok to leave it blank in init templates.
> 
> If not set it, does the api spec still generated but not appear the response info in the api document? ( I would like to check for writing it to the our vulcansql document )

Yes, we'll get an empty response {} in the document.

kokokuo:
> > > > I discover one thing and would like to check does it was missed, The CLI init files `user.yaml` seems not to have `sample` in #86 [code href](https://github.com/Canner/vulcan-sql/pull/86/files#diff-dae705d89fb59e4513ec3d96f19f29bc5c20c67059fca22cb67d8307bd25b44bR9) . Should it also need to add `sample` ?
> 
> > > 
> 
> > > 
> 
> > > Thanks for pointing out, that property is optional, builder won't send sample queries without setting it. So I think it's ok to leave it blank in init templates.
> 
> > 
> 
> > If not set it, does the api spec still generated but not appear the response info in the api document? ( I would like to check for writing it to the our vulcansql document )
> 
> 
> 
> Yes, we'll get an empty response {} in the document. 

Thank you!

kokokuo (approved):
LGTM 👍 👍 👍  
Thanks so much for helping me to develop the feature, really appreciated

PR #91, opened 2022-09-26
Fix: Remove serve package from dependencies list

oscar60310:
## Description

NX detects dependencies from our template js file, so we need to remove it manually.

https://github.com/Canner/vulcan-sql/blob/923605a390b71c81e7871040778362081728c0fa/packages/build/src/models/extensions/packager/assets/entry.js#L1

## Issue ticket number

N/A


kokokuo (approved):
LGTM 👍

PR #95, opened 2022-09-29
fix(serve, cli): fix the checking "format" field logistic in "response-formats" options

### LINKED ISSUE #93: Get Internal Error when not setting "formats" field in "reponse-format" options, and send API request not specify the format in the URL
## Environment

- Version:
  -  cli version: 0.2.1
  - core version: 0.2.1
  - build version: 0.2.1
  - serve version: 0.2.1

- OS: macOS

## What's wrong?
When not setting the `formats` field in `reponse-format` options of the `vulcan.yaml`, the API request will throw an internal error when API URL does not contain any format at the end of the URL.

It should get the `default` format.

## Reproducing steps
1. Don't set the `formats` field in the `reponse-format` options of the `vulcan.yaml`.

```yaml
# vulcan.yaml
response-format:
  enabled: true
  options:
    default: json
```
2. Use a browser to send HTTP GET requests which not has format in the end of the URL
e.g: `http://localhost:3000/api/orders`

3. Got the Internal Error and shows an error message in the terminal:

**Error Response:**
```json
{
"message": "An internal error occurred",
"requestId": "d8d5d068-ef49-4e5b-bc1f-92d97a9308fb"
}
```

**Termal Shows Error**:
` TypeError  Cannot read properties of undefined (reading 'formatToResponse')`

kokokuo:
## Description
- Check the array type when not set `format` field of `response-formats` options and URL does not specify format in the end.
- Update the CLI `vulcan init .` finished message

## Issue ticket number

closes #93

### DIFF
--- a/packages/serve/src/lib/middleware/response-format/helpers.ts
+++ b/packages/serve/src/lib/middleware/response-format/helpers.ts
@@ -29,8 +29,8 @@ export const checkUsableFormat = ({
 
   // if path ending has no format
   if (!pathFormat) {
-    // Use the default when "acceptFormat" us array type, because array type only happened when "supportedFormats" is [] and "Accept" header has multiple
-    if (Array.isArray(acceptFormat)) return defaultFormat;
+    // Use the default format when "supportedFormats" is [].
+    if (supportedFormats.length === 0) return defaultFormat;
     // get default when "Accept" header also not matched or "Accept" header not in request (shows by */*)
     if (!acceptFormat || acceptFormat == '*/*') return defaultFormat;
     // if accept format existed, use "Accept" first matched format by support format order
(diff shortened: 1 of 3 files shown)


packages/serve/src/lib/middleware/response-format/helpers.ts
```diff
@@ -29,6 +29,8 @@ export const checkUsableFormat = ({
 
   // if path ending has no format
   if (!pathFormat) {
+    // Use the default when "acceptFormat" us array type, because array type only happened when "supportedFormats" is [] and "Accept" header has multiple
+    if (Array.isArray(acceptFormat)) return defaultFormat;
```
oscar60310, line 33:
```suggestion
    if (supportedFormats.length === 0) return defaultFormat;
```
Can we check `supportedFormats` instead? When we didn't set any supportedFormat, return the default value. It's more understandable.

oscar60310 (approved):
LGTM

packages/serve/src/lib/middleware/response-format/helpers.ts
kokokuo: Hi @oscar60310, thanks for suggestion, it has been fixed, please check it, thanks

PR #108, opened 2022-10-04
Feature: Syntax sugars for parameter validators

### LINKED ISSUE #99: Syntax sugars for parameter validators

We need to define our parameters in YAML file when we want to add validators or change parameter sources (path, query, or header). But it might be a hassle for data engineers to define parameters in another file.

Provide some syntax sugars in SQL:

Some possible solutions:

1. Functions
   ```sql
   select * from user limit {{ context.params.limit.number({max: 10, min: 0}).required() }}
   ```
  
2. Filters
   ```sql
   select * from user limit {{ context.params.limit | number(max=10, min=0) | required }}
   ```
   We can add a prefix to validators toprevent collision.
   ```sql
   select * from user limit {{ context.params.limit | validate.number(max=10, min=0) | validate.required }}
   ```
   
  ## Additional context

oscar60310:
## Description
I made validation filters for every validator, named with prefix `is_`, e.g. `integer` validator has the `is_integer` filter. Those validators can be used with any variables:

- With dynamic parameters
  ```sql
  SELECT * FROM "artists"
  WHERE ConstituentID = {{ context.params.id | is_required | is_integer }}
  ```

- With predefined queries' result
  ```sql
  {% req artist %}
  SELECT * FROM "artists"
  WHERE DisplayName = {{ context.params.name }} LIMIT 1
  {% endreq %}
  
  SELECT * FROM "artworks"
  WHERE ConstituentID = {{ artist.value()[0].ConstituentID | is_required | is_integer }}
  ```

To add arguments:
```sql
SELECT * FROM users
WHERE age = {{ context.params.age | integer(min=0, max=100) }}
```

Please check the document [here](https://vulcan-sql-document-git-feature-vali-452db1-vulcan-sql-document.vercel.app/docs/api-building/api-validation/validation-filter).

I've also added a new validator `enum` and implemented `getConstraints` function for all validators.

## Issue ticket number

closes #99

## Additional Context
- I used the simplest way to name validation filters: `is_xxxx`, we can improve it in the future.


packages/build/src/lib/schema-parser/middleware/addParameter.ts
```diff
@@ -20,15 +20,18 @@ export class AddParameter extends SchemaParserMiddleware {
...
-          (paramInSchema) => paramInSchema.fieldName === name
-        )
-      ) {
+      const existedParam = schemas.request!.find(
+        (paramInSchema) => paramInSchema.fieldName === name
+      );
+
+      if (existedParam) {
+        if (!existedParam.validators) existedParam.validators = [];
+        existedParam.validators.push(...(parameter.validators || []));
```
kokokuo, line 29: If there are the same name validators in metadata and schema requests, will we exclude one of them? 

Because we may face some cases which the args are different in schema request and SQL syntax validation filter:
- schema request has integer validator with args min=10, and max=20
- metadata exists in the validator with args min=5, and max=90 which parsed from SQL

kokokuo (approved):
Beside one question, other is LGTM 👍 👍

packages/build/src/lib/schema-parser/middleware/addParameter.ts
oscar60310: I pushed all validators to the list because I didn't know how to compose all of the validators, e.g.
string(format="\d+")
string(format="[1-5]+")

So I prefer to execute all of these validators.

P.S. Our constraints have composed logic so that we won't duplicate them.

PR #109, opened 2022-10-04
Fix: Catch all errors to prevent server crashes

### LINKED ISSUE #102: Catch all errors to prevent server crashes

When some uncatched errors are thrown from downstream of Vulcan server, e.g. data source, Vulcan sometimes crashes and exits.

We should catch all errors and keep the server listening.

## Additional context
PostgreSQL data source throws uncached errors when databases disconnected, this issue will be fixed in the future
.

kokokuo:
## Description

- Catch all unexpected errors and keep the server still  listening
   - using `process.on('uncaughtException', (err) => {....})` to catch, including event emitter errors.
- Ignore and warn the connection failed message when executing query in pg driver

## Issue ticket number

closes #102

### DIFF
--- a/packages/extension-driver-pg/src/lib/pgDataSource.ts
+++ b/packages/extension-driver-pg/src/lib/pgDataSource.ts
@@ -54,6 +54,13 @@ export class PGDataSource extends DataSource<any, PGOptions> {
     const { pool, options } = this.poolMapping.get(profileName)!;
     this.logger.debug(`Acquiring connection from ${profileName}`);
     const client = await pool.connect();
+    // listen the pg connection event errors when executing query
+    pool.on('error', (err) => {
+      //  Ignore the error and display warn message
+      this.logger.warn(
+        `Pool client of profile instance ${profileName} connecting failed, detail error, ${err}`
+      );
+    });
     this.logger.debug(`Acquired connection from ${profileName}`);
     try {
       const cursor = client.query(
--- a/packages/serve/src/lib/server.ts
+++ b/packages/serve/src/lib/server.ts
@@ -12,6 +12,7 @@ import {
   ArtifactBuilder,
   InternalError,
   ConfigurationError,
+  VulcanError,
 } from '@vulcan-sql/core';
 import { Container, TYPES } from '../containers';
 import { ServeConfig, sslFileOptions } from '../models';
@@ -87,6 +88,7 @@ export class VulcanServer {
     this.servers = this.runServer(app);
     return this.servers;
   }
+
   public async close() {
     if (this.servers) {
       if (this.servers['http']) this.servers['http'].close();
@@ -146,3 +148,12 @@ export class VulcanServer {
       .listen(httpsPort);
   }
 }
+
+// Listen the all uncaught errors (including event emitter errors) to prevent server stop.
+process.on('uncaughtException', async (err) => {
+  // Display non vulcan error
+  if (!(err instanceof VulcanError)) {
+    logger.warn('Unexpected error happened', err);
+    logger.debug('Make server keep listening');
+  }
+});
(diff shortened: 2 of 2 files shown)


packages/extension-driver-pg/src/lib/pgDataSource.ts
```diff
@@ -54,6 +54,13 @@ export class PGDataSource extends DataSource<any, PGOptions> {
     const { pool, options } = this.poolMapping.get(profileName)!;
     this.logger.debug(`Acquiring connection from ${profileName}`);
     const client = await pool.connect();
+    // listen the pg connection event errors when executing query
+    pool.on('error', (err) => {
+      //  Ignore the error and display warn message
+      this.logger.warn(
+        `Pool client of profile instance ${profileName} connecting failed, detail error, ${err}`
+      );
+    });
```
oscar60310, line 63: NIT: Can we add some test cases for pool issues? e.g. Delete the docker container before connections closed.

oscar60310 (approved):
LGTM, thanks for fixing PostgreSQL issues.

packages/serve/src/lib/server.ts
```diff
@@ -146,3 +148,11 @@ export class VulcanServer {
...
   }
 }
+
+// Listen the all uncaught errors (including event emitter errors) to prevent server stop.
+process.on('uncaughtException', async (err) => {
+  // Display non vulcan error
+  if (!(err instanceof VulcanError))
+    logger.warn('Unexpected error happened', err);
+  logger.debug('Make server keep listening');
+});
```
oscar60310, line 158: We should add the listener when the server started and remove it when the server closed. or everyone who imports anything from `@vulcan-sql/serve` gets this listener.

kokokuo: Hi @oscar60310, thanks for suggestion, it's fixed with [a8fe967](https://github.com/Canner/vulcan-sql/pull/109/commits/a8fe9671df3ec9591f5a80ee067f9bb1b2539e77)

packages/extension-driver-pg/src/lib/pgDataSource.ts
kokokuo: Hi @oscar60310, thanks for the suggestion, I have tried to add the test case by deleting the docker container before connections closed, but seems the test case could not assert the listener work of `pool.on('error', ...)`, or even I try to assert the `pool.emit('error', ...)`, it still not check.

kokokuo: Here is my test code:
```ts

it('Data source should throw error when PG Server disconnected', async () => {
  // Arrange
  const pg = new PGServer();
  await pg.prepare();

  const dataSource = new PGDataSource({}, '', [
    {
      name: 'profile1',
      type: 'pg',
      connection: {
        host: pg.host,
        user: pg.user,
        password: pg.password,
        database: pg.database,
        port: pg.port,
      } as PGOptions,
      allow: '*',
    },
  ]);
  await dataSource.activate();

  const pool = dataSource.getPool('profile1');
  const spyEmit = sinon.default.spy(pool.emit);
  const spyOn = sinon.default.spy(pool.on);

  await pg.destroy();
  // Act
  const executeAction = dataSource.execute({
    statement: 'select * from users limit 1',
    bindParams: new Map(),
    profileName: 'profile1',
    operations: {} as any,
  });

  // Assert
  await expect(executeAction).rejects.toThrow();
  // Assert Pool Emit and On
  expect(spyEmit.calledWith('error', expect.any(Function))).toBeTruthy();
  expect(spyOn.calledWith('error' as any, expect.any(Function))).toBeTruthy();

  await dataSource.destroy();
  // afterEach hook will timeout if any leak occurred.
}, 150000);

```

PR #119, opened 2022-10-19
Feature: Serialize/Deserialize all artifacts

### LINKED ISSUE #105: Sync with catalog server APIs

Catalog server APIs might need some additional information about our endpoints that are not built into artifacts, e.g. request constraints. We should add the missing metadata.

- [x] Serialize/Deserialize request constraints to artifacts. #119

oscar60310:
## Description

The catalog server needs the constraints of each request parameter, but we haven't serialized all of them.
This PR used `class-transformer` to transform classes into plain objects before saving artifacts and to transform plain objects into classes while reading them.

## Issue ticket number

closes #105

## Breaking change
We'd need to rebuild to get constraints from artifacts.

## Additional Context

Here are some simple codes the get constraints. cc @wwwy3y3 
```ts
import * as jsYAML from 'js-yaml';
import { promises as fs } from 'fs';
import * as path from 'path';
import {
  APISchema,
  ArtifactBuilder,
  BuiltInArtifactKeys,
  Container,
  EnumConstraint,
  TYPES,
} from '@vulcan-sql/core';


const container = new Container();

// Read config
const configPath = path.resolve(process.cwd(), 'vulcan.yaml');
const config: any = jsYAML.load(await fs.readFile(configPath, 'utf-8'));

await container.load(config);

const artifactBuilder = container.get<ArtifactBuilder>(TYPES.ArtifactBuilder);
await artifactBuilder.load();

const schemas = artifactBuilder.getArtifact<APISchema[]>(
  BuiltInArtifactKeys.Schemas
);

schemas[0].request[0].constraints.forEach((constraint) => {
  if (constraint instanceof EnumConstraint) {
    console.log(constraint.getList()); // Get the list of enum constraint
  }
});

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,11 @@
   "private": true,
   "dependencies": {
     "@koa/cors": "^3.3.0",
-    "chokidar": "^3.5.3",
-    "bluebird": "^3.7.2",
     "bcryptjs": "^2.4.3",
+    "bluebird": "^3.7.2",
     "bytes": "^3.1.2",
+    "chokidar": "^3.5.3",
+    "class-transformer": "^0.5.1",
     "class-validator": "^0.13.2",
     "commander": "^9.4.0",
     "dayjs": "^1.11.2",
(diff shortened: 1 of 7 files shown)


kokokuo (approved):
LGTM 👍

PR #120, opened 2022-10-21
Fix: Improve the stabilities of PostgreSQL and DuckDB drivers

### LINKED ISSUE #106: Improve the stabilities of PostgreSQL and DuckDB drivers

We found some issues with PostgreSQL and DuckDB drivers, we should fix them.

- [x] Handle all events emitted from node-postgres (#109 have handled some events)
- [x] Bump DuckDB to stable version.

oscar60310:
## Description
PG
- Remove the duplicated event register because it adds listeners for each request but never removes them.
- Fix duplicated connection pools issue.
- Add test cases for error handlers.

DuckDB
- upgrade to 0.5.1

## Issue ticket number

closes #106

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "class-validator": "^0.13.2",
     "commander": "^9.4.0",
     "dayjs": "^1.11.2",
-    "duckdb": "0.4.1-dev2371.0",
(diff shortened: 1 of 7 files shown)


kokokuo (approved):
LGTM 👍

PR #140, opened 2023-01-11
Feature: Dynamic data masking - custom string

### LINKED ISSUE #139: Dynamic data masking - Custom String

We can generate different queries depending on the user attribute, so it is easy to mask some columns manually like below:
```sql
SELECT
  {% if context.user.attr.group == 'admin' %}
    id
  {% else %}
    CONCAT(SUBSTR(id, 0, 4), 'xxxxxx')
  {% endif %} as id,
  age
FROM users;
```
But manually masking each column might be annoying and unpredictable, we should provide "masking" tag with some masking functions.

Inspired by [MSSQL DDM](https://learn.microsoft.com/en-us/sql/relational-databases/security/dynamic-data-masking?view=sql-server-ver16), we can add a built-in tag `masking` with syntax:
```
{% masking <column-name> <masking-function> %}
```

We can implement the most commonly used masking function `partial` first.
For example, assuming that we have a `user` table:
|id|name|
|---|---|
|AB1234567CD|Ivan|
|EF2345678GH|Freda|

```sql
SELECT
  {% masking id partial(2, 'xxxxxxx', 2) %} as id,
  name
FROM user;
```
We'll get the result below:
|id|name|
|---|---|
|ABxxxxxxxCD|Ivan|
|EFxxxxxxxGH|Freda|

fredalai:
## Description

I made the masking tag, and support the partial masking function.

The built-in tag `masking` is added using the following syntax:
```
{% masking <column-name> <masking-function> %}
```

For an example of SQL
```sql
SELECT
  {% masking id partial(2, 'xxxxxxx', 0) %} as id,
  name
FROM user;
```
```sql
SELECT
  {% masking id partial(2, 'xxxxxxx', 2) %} as id,
  name
FROM user;
```

## Issue ticket number

closes #139

## test (yarn run jest masking)
螢幕快照 2023-01-13 下午3 31 46

## Additional Context

 - Implementing the partial function is based on considering the currently supported connectors and using a generic function to implement it.
 - Per discussion with Ivan before, maybe... in order to cope with different connectors, it may be possible in the future that implement a masking function for each connector

### DIFF
--- a/packages/core/src/lib/template-engine/built-in-extensions/masking/maskingTagBuilder.ts
+++ b/packages/core/src/lib/template-engine/built-in-extensions/masking/maskingTagBuilder.ts
@@ -1,0 +1,115 @@
+import * as nunjucks from 'nunjucks';
+import { TagBuilder, VulcanInternalExtension } from '@vulcan-sql/core/models';
+import { SANITIZER_NAME } from '../query-builder/constants';
+
+@VulcanInternalExtension()
+export class MaskingTagBuilder extends TagBuilder {
+  public tags = ['masking'];
+
+  public parse(
+    parser: nunjucks.parser.Parser,
+    nodes: typeof nunjucks.nodes,
+    lexer: typeof nunjucks.lexer
+  ) {
+    // {% masking <column-name> <masking-function> %}
+    const token = parser.nextToken();
+    const columnName = parser.nextToken();
+    const functionName = parser.nextToken();
+    if (functionName.type === lexer.TOKEN_BLOCK_END) {
+      parser.fail(
+        `Expected a symbol, but got a block end`,
+        functionName.lineno,
+        functionName.colno
+      );
+    }
+
+    const argsNodeToPass = new nodes.NodeList(token.lineno, token.colno);
+    argsNodeToPass.addChild(
+      new nodes.Literal(columnName.lineno, columnName.colno, columnName.value)
+    );
+    switch (functionName.value) {
+      case 'partial': {
+        return this.parsePartial(parser, nodes, lexer, argsNodeToPass);
+      }
+
+      default:
+        parser.fail(
+          `Unknown function: ${functionName.value}`,
+          functionName.lineno,
+          functionName.colno
+        );
+    }
+  }
+
+  private parsePartial(
+    parser: nunjucks.parser.Parser,
+    nodes: typeof nunjucks.nodes,
+    lexer: typeof nunjucks.lexer,
+    argsNodeToPass: nunjucks.nodes.NodeList
+  ) {
+    const leftParen = parser.nextToken();
+    if (leftParen.type !== lexer.TOKEN_LEFT_PAREN) {
+      parser.fail(
+        `Expected a function start`,
+        leftParen.lineno,
+        leftParen.colno
+      );
+    }
+
+    const prefix = parser.parseExpression();
+    const commaOne = parser.nextToken();
+    if (commaOne.type !== lexer.TOKEN_COMMA) {
+      parser.fail(
+        `Expected a function has 3 arguments, but got 1`,
+        commaOne.lineno,
+        commaOne.colno
+      );
+    }
+
+    const padding = parser.parseExpression();
+    const commaTwo = parser.nextToken();
+    const commaCount = [commaOne, commaTwo].filter(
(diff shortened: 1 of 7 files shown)


packages/core/src/lib/template-engine/built-in-extensions/masking/maskingTagRunner.ts
```diff
@@ -0,0 +1,31 @@
...
+      return new nunjucks.runtime.SafeString(
+        `concat(${prefixSql}, ${padding})`
+      );
+    }
+
+    // Consider the currently supported connector, using a general function to implement it.
+    const pos = Number(suffix) - 1;
+    return new nunjucks.runtime.SafeString(
+      `concat(${prefixSql}, ${padding}, substr(${columnName}, length(${columnName}) - ${pos}, ${suffix}))`
+    );
```
oscar60310, line 29: We should handle the strings which are too short, for example:

|Data|Masking function|Result|
|------|-------|--------|
|freda|partial(1,'xxx', 1)|fxxxa|
|freda|partial(2,'xxx',2)|frxxxda|
|freda|partial(2,'xxx',3)|xxx|

packages/core/test/template-engine/built-in-extensions/masking/masking.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+it('If not correct use the masking function whole syntax, the extension should throw', async () => {
+  // Arrange
+  const { compiler } = await createTestCompiler();
+
+  // Action, Assert
+  await expect(
+    compiler.compile(`{% masking id partial(2, 'xxxxxxx', 3 %}`)
+  ).rejects.toThrow(`Expected a function end`);
+});
```
oscar60310, line 120: NIT: We can add some test cases for parser position checking:
```sql
AA{% masking id partial(2, 'xxxxxxx', 0) %}AA
```

NIT: We can add some test cases for the toleration of spaces
```sql
{% masking id partial(2,                'xxxxxxx'                   , 3) %}
```

oscar60310 (commented):
Other parts LGTM.

packages/core/src/lib/template-engine/built-in-extensions/masking/maskingTagBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    const [prefix, , padding, , suffix] = functionNextTokens;
+    if (prefix.type !== lexer.TOKEN_INT) {
+      parser.fail(
+        `Expected a int, but got ${prefix.type}`,
+        prefix.lineno,
+        prefix.colno
+      );
+    }
```
oscar60310, line 77: NIT: We might support dynamic values in the future.
```sql
{% set a = 3 %}

select 
ConstituentID,
DisplayName,
{% masking DisplayName partial(a, 'xxxx', 2) %} as DisaplyNameMasked,
from "artists"
limit 10

```

packages/core/src/lib/template-engine/built-in-extensions/masking/maskingTagBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+        functionName.colno
+      );
+    }
+
+    const functionNextTokens = [
+      parser.nextToken(),
+      parser.nextToken(),
+      parser.nextToken(),
+      parser.nextToken(),
+      parser.nextToken(),
```
oscar60310, line 47: We can use parser.parseExpression() instead of reading tokens.

packages/core/src/lib/template-engine/built-in-extensions/masking/maskingTagBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+        suffix.colno
+      );
+    }
+
+    parser.advanceAfterBlockEnd(rightParen.value);
+
+    const argsNodeToPass = new nodes.NodeList(token.lineno, token.colno);
+    argsNodeToPass.addChild(
+      new nodes.Literal(columnName.lineno, columnName.colno, columnName.value)
+    );
```
oscar60310, line 100: NIT: We can add sanitize filter to prvent injection.

oscar60310 (approved):
LGTM

kokokuo (approved):
LGTM 👍

PR #142, opened 2023-01-30
Feature: Passing route information to template engine

### LINKED ISSUE #62: Passing route information to template engine
The route information like host, path ...etc. are useful while rendering SQL queries.

Pass host name and url path into built-in context.
For example: https://example.com/v1/users
{{ context.host }} -> example.com
{{ context.path }} -> /v1/users
or
{{ context.req.host }} ...

## **Additional context**

JSYOU:
## Description

The route information like host, path ...etc. are useful while rendering SQL queries.

```sql
select
  *
from
  "artists"
where
  {{ context.req.method }} = "GET"
and
  {{ context.header.customParameters }} = "custom parameters"
  ......
```

## Issue ticket number

closes #62

### DIFF
--- a/packages/doc/docs/api-building/sql-syntax.mdx
+++ b/packages/doc/docs/api-building/sql-syntax.mdx
@@ -162,9 +162,9 @@ select
 from
   "artists"
 where
-  {{ context.req.method }} = "GET"
+  {{ context.req.method }} = 'GET'
 and
-  {{ context.header.customParameters }} = "custom parameters"
+  {{ context.req.header.customParameters }} = 'custom parameters'
   ......
 ```
 
--- a/packages/doc/docs/api-building/sql-syntax.mdx
+++ b/packages/doc/docs/api-building/sql-syntax.mdx
@@ -150,6 +150,26 @@ from app_data.payments
 
 VulcanSQL will provide the feature by enhancing the [nunjucks](https://mozilla.github.io/nunjucks/templating.html#tags) macro function by importing a macro function from other files in the next version 😃
 
+## Useing HTTP Request data in SQL file
+
+You can easily use `{{ context.req.xxx }}` in the sql file to obtain data from the http request or a custom header as a query condition.
+
+the syntax like the below:
+
+```sql
+select
+  *
+from
+  "artists"
+where
+  {{ context.req.method }} = "GET"
+and
+  {{ context.header.customParameters }} = "custom parameters"
+  ......
+```
+
+For more available parameters, please refer to: [Koajs#Request](https://koajs.com/#request)
+
 ## Set variable
 
 You could set the variable not only primitive value e.g: integer, string, array. But also VulcanSQL support declares a variable by assigning the dynamic parameter to it. Use the `set` tag and `{% ... %}` to declare the variable and set:
--- a/packages/build/src/lib/schema-parser/middleware/responseSampler.ts
+++ b/packages/build/src/lib/schema-parser/middleware/responseSampler.ts
@@ -37,6 +37,7 @@ export class ResponseSampler extends SchemaParserMiddleware {
       {
         parameters: schema.sample.parameters,
         profileName: schema.sample.profile,
+        req: schema.sample?.req,
       },
       // We only need the columns of this query, so we set offset=0 and limit=1 here.
       // Some drivers guess the column types by data, so we should at least query 1 rows.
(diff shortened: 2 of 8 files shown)


packages/doc/docs/api-building/sql-syntax.mdx
```diff
@@ -150,6 +150,26 @@ from app_data.payments
...
+from
+  "artists"
+where
+  {{ context.req.method }} = "GET"
+and
+  {{ context.header.customParameters }} = "custom parameters"
+  ......
+```
+
+For more available parameters, please refer to: [Koajs#Request](https://koajs.com/#request)
```
kokokuo, line 171: Another suggestion is that a sample for using `sample` in the api schema file, because if user would like to make the build time work with sample, he should also add the `req` to sample:

```sql
-- We use the req in sql file
select 
*,
{{ context.req.method }} as "req-method",
{{ context.req.url }} as "req-url"
from "artists"
where 
ConstituentID = {{ context.params.id }}
and
 {{ context.req.method }} = 'GET'

```

```ts
# API Schema
urlPath: /artist/:id
request:
  - fieldName: id
    fieldIn: path
    description: constituent id
    validators:
      - required
sample:
  req:
    method: GET
    url: 'localhost:3000'
  parameters:
    id: '1'
  profile: duck
profile: duck

``

**Review by kokokuo — COMMENTED** [review]
Besides some suggestions Others LGTM 👍

**Comment by kokokuo** [comment]
I think we could update the `labs` sample for adding the `context.req` to make our other members know how to use it and also be a regression test sample when running `labs`. Otherwise, I think we could also update the package version to `[0.3.1-dev.20230205.0](https://www.npmjs.com/package/@vulcan-sql/extension-driver-duckdb/v/0.3.1-dev.20230205.0)`, like below:


```json
{
  "name": "my-first-vulcan-project",
  "dependencies": {
    "@vulcan-sql/core": "0.3.1-dev.20230205.0",
    "@vulcan-sql/extension-driver-duckdb": "0.3.1-dev.20230205.0",
    "@vulcan-sql/serve": "0.3.1-dev.20230205.0"
  },
  "devDependencies": {
    "@vulcan-sql/build": "0.3.1-dev.20230205.0"
  }
}

```

wwwy3y3:
> Otherwise, I think we could also update the package version to [0.3.1-dev.20230205.0](https://www.npmjs.com/package/@vulcan-sql/extension-driver-duckdb/v/0.3.1-dev.20230205.0), like below:

Since it's possible users will just copy our example code in `labs`, maybe we should use the latest stable version (which is `0.3.1` at this moment) ?

kokokuo:
> > Otherwise, I think we could also update the package version to [0.3.1-dev.20230205.0](https://www.npmjs.com/package/@vulcan-sql/extension-driver-duckdb/v/0.3.1-dev.20230205.0), like below:
> 
> Since it's possible users will just copy our example code in `labs`, maybe we should use the latest stable version (which is `0.3.1` at this moment) ?

 

> > Otherwise, I think we could also update the package version to [0.3.1-dev.20230205.0](https://www.npmjs.com/package/@vulcan-sql/extension-driver-duckdb/v/0.3.1-dev.20230205.0), like below:
> 
> Since it's possible users will just copy our example code in `labs`, maybe we should use the latest stable version (which is `0.3.1` at this moment)?

Sure, we could change the stable version `0.3.1`.

But actually, we also provide the [vulcan-demo](https://github.com/Canner/vulcan-demo) to users for playing, and the `labs` actually do not need the package versions ( I recap it after your feedback on the suggestion XD", so actually it doesn't need to update ), because we use the `Makefile` to copy the current developing environment `node_modules` to `labs` for testing.

packages/doc/docs/api-building/sql-syntax.mdx
JSYOU: Hi, @kokokuo , I found that there is already a section about [set-sampler](https://vulcansql.com/docs/api-building/api-document#set-sampler) in the document. Maybe we can update this section to inform the user on how to properly configure it.

kokokuo: Could you open an issue to `record` the part ?

kokokuo: Thanks @JSYOU adding the issue with sample, I record at here https://github.com/Canner/vulcan-sql/issues/146

kokokuo (approved):
LGTM 👍

Issue #143, opened 2023-02-03
Bigquery connection failing


Navrooz:
I have been trying to connect VulcanSQL with BigQuery but receive an error 'Not found'. 

Also, not getting any endpoints in /localhost/doc
![image](https://user-images.githubusercontent.com/4967042/216644022-d6bfbdff-ba73-4591-a8ec-c6fc9cb04444.png)

![image](https://user-images.githubusercontent.com/4967042/216644175-31242919-65d9-4f31-af3b-1dd3f39493f3.png)

The credentials file is working fine as I tested the connection with DBeaver.

- profiles, vulcan, sql and yaml files are at the below link.

Files: https://drive.google.com/drive/folders/1g1lyEG32x63hAzrIo2JK8vFAk5GpNC1U?usp=sharing

wwwy3y3:
hi @Navrooz , thanks for your issue report. We'll check out your attached files and see what's going on.

JSYOU:
Hi @Navrooz, based on your configuration in `vulcan.yaml`, the files `user.sql` and `user.yaml` need to be placed in the `sqls` folder in order for it to run correctly. 

<img width="360" alt="image" src="https://user-images.githubusercontent.com/19495220/216757257-3a863e8f-2eae-4adb-965f-1f077c12468b.png">

Please create a `sqls` folder and move` user.sql` and `user.yaml` into it, then try again. Thank you.

<img width="280" alt="image" src="https://user-images.githubusercontent.com/19495220/216757545-8ff92357-cf9d-4568-93d3-be8da148cb56.png">

Navrooz:
@JSYOU -Hi, Thank you for your reply !
Both the user.sql and user.yaml files are already in sqls folder.
![vulcanTest-1](https://user-images.githubusercontent.com/4967042/216856729-c341c6cf-d0ad-4ee1-9e5c-16d3e2149e4c.png)

JSYOU:
Hi @Navrooz ,
Have you successfully executed your project?

Navrooz:
I tried but as I shared in the screenshots, I get an error 'Not Found' and blank 'localhost:3000/doc' page. It seems like Vulcan is not parsing the endpoint.

JSYOU:
Hi @Navrooz ,
Are you using the `vulcan start` command in the `vulcanTest-1` directory?

We recommend that the working directory be in the same location as the `vulcan.yaml`.

[...]

PR #145, opened 2023-02-09
Fix: Fix the path of sql files not being recognized correctly in winOS

JSYOU:
## Description

Fix the path of sql files not being recognized correctly in winOS
## Issue ticket number
#143 

## Additional Context
Windows use backslash when using path.resolve, e.g. `C:\\Users\\xxxxx\\cli`, but `glob` accepts only forward ward slash, so we make the windows path to linux path to make `glob` accepted

Glob: https://github.com/isaacs/node-glob#windows
Path separator: https://nodejs.org/api/path.html#pathsep


kokokuo (approved):
LGTM 👍

Issue #154, opened 2023-04-06
Define cache layer loader to load parquet files to in-memory duckDB 


kokokuo:
## What’s the problem you're trying to solve
In #150, In order to enhance our query performance after users send the API request to run our data endpoint to get the result from the data source. We need to provide a **Caching (pre-loading) Layer with the duckDB** to enhance query performance.

## Describe the solution you’d like
In order to send the query result from a data source and keep in the cache layer with duckDB after users define the cache config in schema YAML,  we need a method to keep the query result and pass it to duckDB by creating the table. 

### Export to parquet format
The method is to keep the query result in a [Parquet](https://parquet.apache.org/) format file, the reason is that we could make the result portable and put it in a local or remote place and share it with other users to use.

The directory format is `[templateName]/[profileName]/[cachedTableName]` and each data source create it own file name *(maybe a lot of partition file) `[folderPartitionName].parquet` and put in the directory.

### The time to export result to parquet and load to duckDB
Otherwise, we decide to send the query result from a data source and keep it to parquet,  and when to read the parquet files and load them to duckDB when users type `vulcan serve` to start our vulcan server. The reason is to prevent making the workload happen when API receives:

- If not find any parquet files in the `[templateName]/[profileName]/[cachedTableName]` directory :
    1. Read data sources data and export to parquet 
    2. Then read Parquet to load data to duckDB.
- If exist parquet files in the `[templateName]/[profileName]/[cachedTableName]` direcotry,  check if  `refreshExpression` or `addRefreshTime` is set and triggered and do the refresh step:
    1. Read data sources data and export to parquet
    2. Then read Parquet to load data to duckDB.

### Define CacheLayerLoader to control the flow
Here we define `CacheLayerLoader` class to implement delegating data sources export query results to parquet files according to artifact, and then get the parquet file and load parquet to duckDB. 
- The `CacheLayerLoader` injects all `DataSource` when the container is loading.
- Define a method named `preload` :
    - The `preload` calls the `export` from each `DataSource` and keeps the location and in key mapper to keep `cachedTableName` and `[templateName]/[profileName]/[cachedTableName]` directory. Then reading all parquets in the `[templateName]/[profileName]/[cachedTableName]` directory for loading to duckDB is needed.
- Load all parquet files data in the `[templateName]/[profileName]/[cachedTableName]` directory and create `cacheTableName` in duckDB by `preload` of `CacheLayerLoader` after all parquet exported.

## Additional Context
[...]

Issue #153, opened 2023-04-06
Support `cache` tag could keep result query result to variable like `req`


kokokuo:
## What’s the problem you're trying to solve
In #150 In order to enhance our query performance after users send the API request to run our data endpoint to get the result from the data source. We need to provide a Caching (pre-loading) Layer with the duckDB to enhance query performance.

## Describe the solution you’d like
We also support   `{% cache %} … {% endcache %}` tag could keep the result from duckDB cache to variable and reuse like `req` tag

```sql
-- omit the setting cache information in YAML by cache key or {% schema set(...) %} in SQL

-- keep the result from {% cache %} ... {% endcache %} and assign to the variable named "prices_range_orders"
# return the result by send the query to duckDB cache data source of the below SQL
{% cache prices_range_orders %}
  select * from order
    where type = {{ context.params.type }}
    and where exists (
        select * from product where price >= {{ context.params.price }}
	and order.product_id = product.id
   )
{% endcache %}

# req main
select * from public.users 
where users.id = {{ prices_range_orders.value()[0].buyer_id }}
```

Issue #151, opened 2023-04-06
Parse `cache` config from API Schema YAML


kokokuo:
## What’s the problem you're trying to solve
In #150, In order to enhance our query performance after users send the API request to run our data endpoint to get the result from the data source. We need to provide a **Caching (pre-loading) Layer with the duckDB** to enhance query performance.

## Describe the solution you’d like
In the part, only focus on parsing the YAML by the above `cache` format.

```yaml
# API schema YAML
cache:
  # "order" table keep in duckDB
  - cacheTableName: "order"
      sql: "select * from order",
      # optional
      profile: pg
      # optional
      refreshTime:
        every: "5m"
      # optional
      refreshExpression:
        expression: "MAX(created_at)"
        every: "5m"
      # optional
      indexes:
        'idx_a': 'col_a'
        'idx_b': 'col_b'

  # "product" table  keep in duckDB
  - cacheTableName: "product"
       sql: "select * from product",
```

If the `refreshExpression` and `refreshTime` are both setup, throw the error when the user type `vulcan build`.

## Additional Context
Because we will solve the **Caching (pre-loading) Layer with the duckDB** by multiple issues, we need to define an interface to interact with the solution of other issues.

Here, make the `APISchema` like below, we add the  `cache!: Array<CacheLayerInfo>;` in `APISchema` and define the `CacheLayerInfo` interface. 
```ts
// packages/core/src/models/artifact.ts

export class CacheLayerInfo {
  cacheTableName!: string;
  sql!: string;
  refreshTime?: Record<string, string>;
  refreshExpression?: Record<string, string>;
  // the used profile to query the data from data source
  profile!: string;
  // index key name -> index column
  indexes?: Record<string, string>;
}

export class APISchema {
  // graphql operation name
  operationName!: string;
  // restful url path
  urlPath!: string;
  // template, could be name or path
  templateSource!: string;
  @Type(() => RequestSchema)
  request!: Array<RequestSchema>;
  @Type(() => ErrorInfo)
  errors!: Array<ErrorInfo>;
  @Type(() => ResponseProperty)
  response!: Array<ResponseProperty>;
  description?: string;
  // The pagination strategy that do paginate when querying
  // If not set pagination, then API request not provide the field to do it
  pagination?: PaginationSchema;
  sample?: Sample;
  profiles!: Array<string>;
  cache!: Array<CacheLayerInfo>;
}
```

[...]

PR #162, opened 2023-04-24
Feature: Support SQL syntax of `cache` tag with keeping to variable like `req` and cache layer loader

### LINKED ISSUE #152: Support SQL syntax of `{% cache %} ... {% endcache %}` tag
In #150, In order to enhance our query performance after users send the API request to run our data endpoint to get the result from the data source. We need to provide a **Caching (pre-loading) Layer with the duckDB** to enhance query performance.

Define the tag scope of `{% cache %} ... {% endcache %}` to parse the scope of `{% cache %} ... {% endcache %}` tag and send the SQL to duckDB. 

```sql
-- The "cache" scope means the SQL statement will send the query to duckDB cache data source
{% cache %}
select * from order
  where type = {{ context.params.type }}
  and where exists (
    select * from product  where price >= {{ context.params.price }}
    and order.product_id = product.id
  )
{% endcache %}
```

## Additional Context
In the build time, define a `CacheTagBuilder` to parse `{% cache %} … {% endcache %}` to the AST tree, set the cache metadata name `cache.vulcan.com` to make schema parser #151 to work, then keep to the artifact.

kokokuo:
### Support `cache` tag & cache layer loader
Covered by 4cbfaf3, d9f9db4, c0ca9fa, 55160d4, 07ff6d2, 11864c8, e61bbb9.
Support the `{% cache %} ... {% endcache %}` tag to make the user could query the result from the cache layer storage like below:

```sql
{% cache %}
select * from cache_departments
where "name" = {{ context.params.name }};
{% endcache %}
```

Users could define which the query from data source preload to the cache layer according to the cache setting in the API schema

 ```yaml
urlPath: /departments
request:
  - fieldName: name
    fieldIn: query
    description: department name
    validators:
      - string
      - required
profile: pg
cache:
  - cacheTableName: 'cache_departments'
    sql: 'select * from departments'
    profile: pg
```

The `cache_departments` will be used in  `{% cache %} ... {% endcache %}`.

The cache layer used storage according to the `loader` field of `cache` setting in `vulcan.yaml`, currently supports duckdDB be cache layer, in the future, we possibly support other storage be our cache layer e.g: click house.

```yaml
# vulcan.yaml
...

cache:
  type: parquet
  folder: tmp
  loader: duckdb
```

Above you will see we also need to set `type` and `folder`, the reason is that vulcan will export the query from the data source to the indicated file format ( currently only support parquet ), and reload to the cache layer. It could make us reuse or refresh the data to cache the layer.

### Support keeping the query result from the cache layer to a variable like `req`
Covered by 11864c8, e61bbb9
VulcanSQL also supports making the cache tag query result keep to a variable like `req` and reuse it, like below:

```sql
-- See the example in "integration-testing" package
{% cache employee %}
select * from cache_employees where "id" = {{ context.params.id }};
{% endcache %}

select * from departments where "employee_id" = {{ employee.value()[0].id }};
```

When the `{% cache %}` not set variable, it will get the result from the cache builder directly ( Also named cache main builder ).
It only could use in the final result, and it will replace the original main query builder, so we could use multiple `cache` tag with variables and one `cache` tag without variable like below:

```sql
-- See the example in "integration-testing" package
{% cache stores %}
select * from cache_stores where "id" = {{ context.params.id }};
{% endcache %}

{% cache %}
select * from cache_products where "store_id" = {{ stores.value()[0].id }};
{% endcache %}
```

### Cache tag syntax enhancement 
Covered by e61bbb9, ab436e7
Cache tag syntax enhancement for not allowing duplicate variable names more than on cache tag like below:

```sql
-- Failed > Duplicate variable name
{% cache variable1 %} select * from tableA {% endcache %}
{% cache variable1 %} select * from tableB {% endcache %}

-- Succeed > Different variable name
{% cache variable1 %} select * from tableA {% endcache %}
{% cache variable2 %} select * from tableB {% endcache %}
```

Otherwise, only allow one `cache` tag without the variable name.

```sql
-- Failed > Not allow more than one cache tag without the variable name.
{% cache %} select * from tableA {% endcache %}
{% cache %} select * from tableB {% endcache %}
```

### Not allow the same cache table name exists in more than one API schema
Covered by ab436e7
If the user defines the same cache table name in more thane one API schema, it will be failed like the below `cache_departments`:

```yaml
urlPath: /departments
request:
  - fieldName: name
    fieldIn: query
    description: department name
    validators:
      - string
      - required
profile: duck
cache:
  - cacheTableName: 'cache_departments'
    sql: 'select * from departments'
    profile: duck
```

```yaml
urlPath: /department/:id/employees
request:
  - fieldName: id
    fieldIn: path
    description: employee id
    validators:
      - uuid
      - required
    profile: pg
cache:
  - cacheTableName: 'cache_departments'
    sql: 'select * from departments'
```

## Issue ticket number

Closes #152, #153, #154 

## Additional Context
- We refactor the original create data source factory in `exectuorModule` to a `dataSourceModule`.
- We will support the cache layer feature that could refresh according to `refresh` settings defined from #151 in other PR.
- When users use the cache main builder by cache tag without variable, we will replace the original main builder to cache main builder to get the query result by `.value()` in `execute()` of template engine.
- support importing multiple parquet files to create one table in duckdb by reading the directory at d068f11679d2ef181850599540ff2d0d407d0052, 903f3ff
- updating the cache layer loader to export parquet files to the specific directory and import all parquet files from the directory at 903f3ff
- support giving default temporary folder path if the user did not set it in the "cache" settings of `vulcan.yaml` at d068f11
- Update to create a new connection for querying, exporting, and importing to duckdb at 5f3faa8
- Update the `duckdb.d.ts` at 5f3faa8.


kokokuo:
@oscar60310 could you review my PR 😄 , cc @wwwy3y3

kokokuo:
Updated the code to support @wwwy3y3's suggestion:
- updating the cache layer loader to export parquet files to the specific directory and import all parquet files from the directory at 2a0111e
- support giving default temporary folder path if the user did not set it in the "cache" settings of `vulcan.yaml` at 6ea2913
- Update to create a new connection for querying, exporting, and importing to duckdb at 9422c53
- Update the `duckdb.d.ts` at 9422c53.

packages/core/test/template-engine/built-in-extensions/cache-layer/cache.spec.ts
```diff
@@ -0,0 +1,62 @@
+import { createTestCompiler } from '../../testCompiler';
+
+it('If exist arguments behind "cache" symbol, extension should throw an error', async () => {
+  // Arrange
+  const { compiler } = await createTestCompiler();
+
+  // Action, Assert
+  await expect(
+    compiler.compile(`{% cache variable %} some statement {% endcache %}`)
+  ).rejects.toThrow(`Expected a block end, but got symbol`);
```
onlyjackfrost, line 35: just curious, what will happen if there exist arguments behind the "endcache" symbol

packages/core/test/template-engine/built-in-extensions/cache-layer/cache.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+it('The metadata should include"isUsedTag" is true in "cache.vulcan.com" after compiled', async () => {
+  // Arrange
+  const { compiler } = await createTestCompiler();
+
+  // Action
+  const { metadata } = await compiler.compile(
+    `{% cache %} some statement {% endcache %}`
+  );
+
```
onlyjackfrost, line 109: should we add a test case to indicate that the isUsedTag will be true even if there is no statement body in the cache tag?
``` typescript
const { metadata } = await compiler.compile(
    `{% cache %}{% endcache %}`
  );
await expect(metadata).toMatchObject({
    'cache.vulcan.com': { isUsedTag: true },
  });
```

packages/core/src/lib/template-engine/built-in-extensions/cache-layer/index.ts
```diff
@@ -0,0 +1,3 @@
+import { CacheTagBuilder } from './cacheTagBuilder';
```
onlyjackfrost, line 1: I am a little bit confused about the folder name, cause we have another "cache-layer" folder under /core/src/lib
and we have a "query-builder" folder that seems to work the same as the "template-engine/build-in-extension/cache-layer"

onlyjackfrost: Maybe rename it to "cache-builder"?

fredalai: Maybe `cache`?

fredalai (approved):
LGTM.

Additional note:
- It seems that we can use the cache table set in other .yaml files in a .sql file. I think it's fine now, let's see if we need to deal with it further in the future.
- I think we also need to update the related guide in the document.

packages/core/test/template-engine/built-in-extensions/cache-layer/cache.spec.ts
kokokuo: Thanks for asking the question, it depends on the Nunjucks parser rule, and according to the test case, I add after you asked, looks like it will throw an error caused by Nunjucks.

For making our other members also know the case, I will add the test case!

packages/core/test/template-engine/built-in-extensions/cache-layer/cache.spec.ts
kokokuo: Sure, but I think we don't need to add another test case because it's another sample for testing the `isUsedTag`, so I will use `jest.each` to add the sample it the original test cases

packages/core/src/lib/template-engine/built-in-extensions/cache-layer/index.ts
kokokuo: I think `cache` is good.

packages/core/test/template-engine/built-in-extensions/cache-layer/cache.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+  await expect(
+    compiler.compile(`
+{% cache variable more %}
+select count(*) as count from user where user.id = '{{ params.userId }}';
+{% endcache %}
+  `)
+  ).rejects.toThrow(`Expected a block end, but got symbol`);
+});
+
+it('If argument is not a symbol or end block, extension should throw', async () => {
```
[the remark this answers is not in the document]
kokokuo: After discussion, because the Extension not have the compile method, and it delegates the compiler to parse in the compiling phase, so we change to use `should be compiled failed` to the suitable description

packages/core/test/template-engine/built-in-extensions/cache-layer/cache.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+  // Arrange
+  const { compiler } = await createTestCompiler();
+
+  // Action, Assert
+  await expect(
+    compiler.compile(`{% cache %} some statement {% endcache %}`)
+  ).resolves.not.toThrow();
+});
+
+it('If argument have too many elements, extension should throw an error', async () => {
```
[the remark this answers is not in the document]
kokokuo: Sure, I will also change other similar descriptions.

packages/core/src/lib/cache-layer/cacheLayerLoader.ts
```diff
@@ -0,0 +1,116 @@
...
+          directory,
+          profileName: profile,
+          type: this.options.type!,
+        });
+        // 2. preload the files to cache data source
+        await cacheStorage.import({
+          tableName: cacheTableName,
+          directory,
+          // use the "vulcan.cache" profile to import the cache data
+          profileName: cacheProfileName,
```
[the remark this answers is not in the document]
kokokuo: No, the `profileName` here only accepts the value user set in the profiles, but our `vulcan.cache` is our built-in name actually, so we could pass it here.

packages/core/src/lib/cache-layer/cacheLayerLoader.ts
```diff
@@ -0,0 +1,116 @@
...
+   * Preload the cache files to cache data source and created tables
+   * @param schemas The API schemas from artifact to preload
+   */
+  public async preload(schemas: Array<APISchema>): Promise<void> {
+    // prepare cache data source
+    const cacheStorage = this.dataSourceFactory(cacheProfileName);
+    // check if the cache table name is duplicated more than one API schemas
+    this.checkDuplicateCacheTableName(schemas);
+    // traverse each cache table of each schema
+    for (const schema of schemas) {
```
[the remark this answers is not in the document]
kokokuo: Sure, after using return await Promise directly in `export` and `import` methods, the `import` should be work for waiting the `export` behavior finished even if it's asynchronous.

packages/core/src/lib/cache-layer/cacheLayerLoader.ts
```diff
@@ -0,0 +1,116 @@
...
+          schema: vulcanCacheSchemaName,
+          type: this.options.type!,
+        });
+      }
+    }
+  }
+
+  private checkDuplicateCacheTableName(schemas: APISchema[]) {
+    const tableNames = schemas
+      // => [[table1, table2], [table1, table3], [undefined, table4], undefined]
```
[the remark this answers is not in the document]
kokokuo: Thanks for reviewing, I have removed the undefined value sample in the comment.

packages/core/src/lib/cache-layer/cacheLayerLoader.ts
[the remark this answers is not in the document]
kokokuo: The `vulcan.cache` is only used for cache layer need, however, we still have other data sources, and not all data sources will use the import method to be the cache layer because the import method could be used to import data from parquet files actually, so user or future work may possibly support import method but not use in the cache layer

onlyjackfrost (approved):
LGTM

PR #163, opened 2023-04-24
Feature: implement snowflake export function #156

onlyjackfrost:
## Description

This pull request implemented the snowflake export function to export the selected data to the local dir in parquet format.

export flow:
 - Use the snowflake "[COPY INTO...](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table)" command to convert the selected data into parquet format and stored in the user stage. This request will have a UUID generated by Snowflake, we will append this unique id to the stage file name to prevent concurrent copy requests with the same name which Snowflake will throw an error.
 - Use the snowflake "[GET](https://docs.snowflake.com/en/sql-reference/sql/get) ..." command to download the parquet file stored in the user stage to the local directory. We Use the "pattern" parameter in this command and filter out the file with the unique id.
 - Use the snowflake "[Remove](https://docs.snowflake.com/en/sql-reference/sql/remove) ..." command to remove the parquet file in the user stage to avoid increasing costs. We Use the "pattern" parameter in this command and filter out the file with the unique id.

## Issue ticket number
issue #156  

## Additional Context
 - The encoding type of columns in the parquet file seems can not be specified in Snowflake.
 - We add a query Id (a uuid generated by snowflake) in our stage file name and allow partition when reach the file size limit
 - We set the max file size of each file to the maximum of snowflake setting (5GB)
 - The max file size does not have a lower bound, but the minimum file size is 1 Mb though. (found when developing)
 - the partition file will have a subfix in the file name like "_\d{1,2}_\d{1,2}_\d{1,2}\.parquet"
    
image

 - The COPY command does not validate data type conversions for Parquet files.  see [Usage Note](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#usage-notes).
 - The GET command will auto-decrypt the parquet file with the key that is used when encrypting., see [Usage Note](https://docs.snowflake.com/en/sql-reference/sql/get#usage-notes)
 - Still can not get the proper encoded type in column after adjusting the session parameter [ENABLE_UNLOAD_PHYSICAL_TYPE_OPTIMIZATION](https://docs.snowflake.com/en/sql-reference/parameters#enable-unload-physical-type-optimization) , see [Note](https://docs.snowflake.com/en/user-guide/script-data-load-transform-parquet#step-5-unload-the-table)(May encounter invalid parquet type when reading parquet file using [parquetjs](https://github.com/ironSource/parquetjs) or derivative package like [parquets](https://github.com/kbajalc/parquets))

## Result
Use parquet-cli to view the downloaded parquet file
image


packages/extension-driver-snowflake/src/lib/snowflakeDataSource.ts
```diff
@@ -100,6 +97,78 @@ export class SnowflakeDataSource extends DataSource<any, SnowflakeOptions> {
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    return new Promise((resolve, reject) => {
+      let status = statement.getStatus();
+      while (status === 'fetching') {
+        status = statement.getStatus();
+        if (status === 'complete') {
+          resolve(true);
+        }
+      }
+      resolve(true);
```
kokokuo, line 168: Seems that we do not get the `waitStatement` result to use, and here no `resolve(false)` case need, so we could only `resolve()` directly 😄 

Btw, The [snowflak-sdk](https://github.com/snowflakedb/snowflake-connector-nodejs/blob/4783f0ca2d8f4d859cf4d9a1663818de993c5ef8/lib/connection/statement.js#L531) only has two status: `fetching` and `complete`, so maybe we could only check the while not equal to `'complete'` directly and if the status is `complete`, then it will leave the loop and `resolve()` directly:

```ts
return new Promise((resolve, reject) => {
      const status = statement.getStatus();
      while (status !== StatementStatus.Complete);
      resolve();
    });
```

packages/extension-driver-snowflake/src/lib/snowflakeDataSource.ts
```diff
@@ -111,7 +180,7 @@ export class SnowflakeDataSource extends DataSource<any, SnowflakeOptions> {
         streamResult: true,
         binds: binding,
         complete: (err, stmt) => {
-          if (err) return reject(err);
+          if (err) reject(err);
```
kokokuo, line 183: Thanks for fixing the unnecessary `return` 😄

packages/extension-driver-snowflake/test/snowflakeDataSource.spec.ts
```diff
@@ -1,6 +1,9 @@
 import { SnowflakeServer } from './snowflakeServer';
 import { SnowflakeDataSource } from '../src';
-import { streamToArray } from '@vulcan-sql/core';
+// import { streamToArray } from '@vulcan-sql/core';
```
kokokuo, line 3: Please remove the commented import line, thanks 😄

packages/extension-driver-snowflake/test/snowflakeDataSource.spec.ts
```diff
@@ -56,6 +59,75 @@ it('Data source should return correct rows and types', async () => {
...
+    FROM TABLE(GENERATOR(ROWCOUNT => ${rawCount})) v 
+    ORDER BY 1`,
+    profileName: 'profile1',
+    directory,
+  };
+  await dataSource.export(exportOption);
+  // Assert
+  const files = fs.readdirSync(directory);
+  expect(files.length).toBe(1);
+  expect(/parquet$/.exec(path.basename(files[0])));
```
kokokuo, line 83: Well, the `expect(/parquet$/.exec(path.basename(files[0])))` actually may not really assert the result.

The value in the `expect(value)` is the actual value, we should call the method to do the assert.

Please use the [expect](https://jestjs.io/docs/expect) method to assert the result, e.g:

```ts
expect(files[0]).toMatch('/.parquet$/');
```

packages/extension-driver-snowflake/test/snowflakeDataSource.spec.ts
```diff
@@ -56,6 +59,75 @@ it('Data source should return correct rows and types', async () => {
...
+    ORDER BY 1`,
+    profileName: 'profile1',
+    directory,
+  };
+  await dataSource.export(exportOption);
+  // Assert
+  const files = fs.readdirSync(directory);
+  expect(files.length).toBe(4);
+  // check each file in the files exists
+  files.forEach((file) => {
```
kokokuo, line 123: The same issue and suggestion line above.

packages/extension-driver-snowflake/test/snowflakeDataSource.spec.ts
kokokuo, line 1: Suggest to enhance test cases, according to the `export` method logistics, we may still need two test cases, first is checking the directory not exist and throw error, second is the checking the happened cached error case

packages/extension-driver-snowflake/src/lib/snowflakeDataSource.ts
```diff
@@ -100,6 +97,78 @@ export class SnowflakeDataSource extends DataSource<any, SnowflakeOptions> {
...
+    // throw error if filepath dir not exist
+    if (!fs.existsSync(directory)) {
+      throw new InternalError(`Directory ${directory} not exist`);
+    }
+
+    try {
+      const stage = '@~/vulcan_cache_stage';
+      const stageFilePath = `${stage}/`;
+      const builtSQL = this.getCopyToStageSQL(sql, stageFilePath);
+      // copy data to a named stage
```
kokokuo, line 116: Suggest to rich the comment to explain why we need to copy to the stage to help members know moving to the stage is necessary.

packages/extension-driver-snowflake/src/lib/snowflakeDataSource.ts
```diff
@@ -100,6 +97,78 @@ export class SnowflakeDataSource extends DataSource<any, SnowflakeOptions> {
...
+      const absoluteDir = path.resolve(directory);
+      const getStatement = await this.getStatementFromSQL(
+        connection,
+        `GET ${stage} file://${absoluteDir} pattern='${pattern}'`,
+        []
+      );
+      await this.waitStatement(getStatement);
+      this.logger.debug(`Exported parquet files to ${directory}`);
+
+      // remove parquet file from stage
```
kokokuo, line 137: Suggest rich the comment to describe removing the stage parquet files because we keep them in the stage temporarily for downloading, so it need to remove 😄

kokokuo (commented):
Besides some suggestions, others LGTM 👍

packages/core/src/models/extensions/dataSource.ts
```diff
@@ -7,6 +7,14 @@ import { InternalError } from '../../lib/utils/errors';
 import { ExtensionBase } from './base';
 import { VulcanExtension } from './decorators';
 
+export interface ExportOptions {
+  // The sql query result to export
+  sql: string;
+  // The full pathname to export result to file
+  directory: string;
+  // The profile name to select to export data
+  profileName: string;
```
kokokuo, line 16: Could you assist to update and async the `exportOptions` to prevent solve the conflict when rebase the code, thanks:

```ts
export interface ExportOptions {
  // The sql query result to export
  sql: string;
  // The directory to export result to file
  directory: string;
  // The profile name to select to export data
  profileName: string;
  // export file format type
  type: CacheLayerStoreFormatType | string;
}
```

kokokuo (approved):
Thanks for fixing and update the commits, LGTM 👍 👍

Issue #150, opened 2023-04-06
Epic Request: Caching (pre-loading) Layer with the duckDB


kokokuo:
## What’s the problem you're trying to solve
In order to enhance our query performance after users send the API request to run our data endpoint to get the result from the data source. We need to provide a **Caching (pre-loading) Layer with the duckDB** to enhance query performance.

## Describe the solution you’d like
Define the cache layer in YAML by `cache` field, and define each `cacheTableName` and its config. For example we define an sql  `select * from order` query the result from data source `pg` and keep to duckDB by `order` table name.

Otherwise, the `refreshTime` and `refreshExpression` means when should we need to refresh the query and keep the new result in the cache table in duckDB. The `index` is a [duckDB indexes feature](https://duckdb.org/docs/sql/indexes.html) to index like the original database and make query efficient.

```yaml
# API schema YAML
cache:
  # "order" table keep in duckDB
  - cacheTableName: "order"
      sql: "select * from order",
      # optional
      profile: pg
      # optional
      refreshTime:
        every: "5m"	
      # optional
      refreshExpression:
        expression: "MAX(created_at)"
        every: "5m"
      # optional
      indexes:
        'idx_a': 'col_a'
        'idx_b': 'col_b'

  # "product" table  keep in duckDB
  - cacheTableName: "product"
       sql: "select * from product",

# used data source
profiles: pg
```

Here is our SQL file, which makes our query send to duckDB to get results. In the scop of `{% cache %} ... {% endcache %}` , when we send the query to duckDB, we will search the each table keep in duckDB or not, so if your YAML not defined the same cache table name, it will make the query failed.
```sql
-- The "cache" scope means the SQL statement will send the query to duckDB cache data source
{% cache %}
select * from order
  where type = {{ context.params.type }}
  and where exists (
    select * from product  where price >= {{ context.params.price }}
    and order.product_id = product.id
  )
{% endcache %}
```

## Breakdown
- [x]  #151 
- [x]  #152
- [x]  #153
- [x]  #154
- [x]  #155
- [x]  #156
- [x]  #164

The #157 not contains in the epic, we will arrange to develop it in the future.

PR #173, opened 2023-05-24
Feature: Support indexing column and refreshing data by time the in the cache layer 

### LINKED ISSUE #164: Support indexing column and refreshing data by time the in the cache layer
In https://github.com/Canner/vulcan-sql/issues/150, In order to enhance our query performance after users send the API request to run our data endpoint to get the result from the data source. We need to provide a Caching (pre-loading) Layer with the duckDB to enhance query performance. 

Based on #154, after defining the cache layer loader to support loading parquet files to in-memory duckDB, we also need to enhance two parts:
1. support the cache layer data refresh by our API schema that defined `refrehTime` or `refreshExpression`options in cache settings.
2. support building column index when loading the parquet files data to create cache table according to user set `indexes` options in cache settings or not.

```yaml
# API schema YAML
cache:
  # "order" table keep in duckDB
  - cacheTableName: "order"
      sql: "select * from order",
      # optional
      profile: pg
      # optional
      refreshTime:
        every: "5m"	
      # optional
      indexes:
        'idx_a': 'col_a'
        'idx_b': 'col_b'

  # "product" table  keep in duckDB
  - cacheTableName: "product"
      sql: "select * from product",
      # optional
      refreshExpression:
        expression: "MAX(created_at)"
        every: "5m"
  
# used data source
profiles: pg
```

In this issue, for the first refreshing data part, we focus on supporting `refreshTime` field and refresh according to the time reached since the last time we exported the parquet and loaded it to the cache since.

### Update parquet file format
We update the original parquet files format to add timestamps for making us recognize the last time we start to do the refresh task.
- Original format :
```
# Format
[folderPath]/[templateSource]/[profileName]/[cacheTableName]/[filePartitionName].parquet
# Example
/tmp/schema1/pg/table1/file_0.parquet
/tmp/schema1/pg/table1/file_1.parquet
```
- Add timestamp : 
```
# Format
[folderPath]/[templateSource]/[profileName]/[cacheTableName]/[timestamp]/[filePartitionName].parquet
# Example
/tmp/schema1/pg/table1/1682653052/file_0.parquet
/tmp/schema1/pg/table1/1682653052/file_1.parquet
```

The step to add the timestamp will make `CacheLayerLoader` do, so each data source won't add it in the `export` methods and then call `export` method.

### The workflow
1. We will export all query results from the data source to parquets and load them to create the table before the server start-up.  At this time, we generated our first parquets including the timestamp in the beginning.
2. Start to schedule the workflow that exporting and loading to create a cache table according to every cache setting of API schema set `refreshTime` field.

kokokuo:
In https://github.com/Canner/vulcan-sql/issues/150, In order to enhance our query performance after users send the API request to run our data endpoint to get the result from the data source. We need to provide a Caching (pre-loading) Layer with the duckDB to enhance query performance. 

Based on #154, after defining the cache layer loader to support loading parquet files to in-memory duckDB, we also need to enhance two parts:
1. support the cache layer data refresh by our API schema that defined `refrehTime` or `refreshExpression`options in cache settings ( Note: In this issue only support `refreshTime` options ).
2. support building column index when loading the parquet files data to create cache table according to user set `indexes` options in cache settings or not.

```yaml
# API schema YAML
cache:
  # "order" table keep in duckDB
  - cacheTableName: "order"
      sql: "select * from order",
      # optional
      profile: pg
      # optional
      refreshTime:
        every: "5m"	
      # optional
      indexes:
        'idx_a': 'col_a'
        'idx_b': 'col_b'
  
# used data source
profiles: pg
```
## Issue ticket number

closes #164

## Additional Context
1. The index of the `duckdb` should be unique, so add the validation to the `checkCache` middleware to check it in the same schema, otherwise, check the index naming pattern ( same as the table name ).
2. Based on the above, add the validation to check the index name should not be duplicated by more than one API schema ( in the `CacheLayerRefresher` )
3. We use the `toad-scheduler` package to handle the refresh time setting, if the cache will refresh by 5 seconds, but the task need to handle 6 seconds, then when next time the refresh schedule is triggered and discover the task is still working, the scheduler will skip and wait for next 5 seconds to check again, like below sample:
```bash

scheduler triggered time: 13:00, 13:05, 13:10, 13:15:
- 13:00 > handle task and 13:06 finished the task, so the scheduler will skip the 13:05 trigger event
- 13:10 > handle task and 13:16 finished the task, so the scheduler will skip the 13:15 trigger event
```
4. We use the jest faker timer to test `CacheLayerRefresher`, because our task is a promise task, so we create a `flushPromise` helper function to flush all pending promises in the event loop, please refer: https://gist.github.com/apieceofbart/e6dea8d884d29cf88cdb54ef14ddbcc4
5.  We add `.eslintignore` to prevent linting the sub `node_modules` in the mono repo package, here indicating the `extension-driver-snowflake`, please refer https://github.com/eslint/eslint/discussions/14544
6. Extend the timeout to 30m to prevent CircleCI default timeout (10m) reached and stop the long testing and cause failed. ( Our tests will spent more time follow by ourt test cases increased )


packages/extension-driver-duckdb/src/lib/duckdbDataSource.ts
```diff
@@ -198,9 +197,28 @@ export class DuckDBDataSource extends DataSource<any, DuckDBOptions> {
...
+  // create indexes for table
+  private createIndexes(
+    connection: duckdb.Connection,
+    schema: string,
+    tableName: string,
+    indexes: Record<string, string>
+  ) {
+    Object.entries(indexes).forEach(([index, column]) => {
+      // duckdb could not create duplicate index, so here should drop index if exists and create again.
+      connection.run(`DROP INDEX IF EXISTS ${index} ON ${schema}.${tableName}`);
```
onlyjackfrost, line 218: Could you tell me the purpose of recreating an index with the same name on the same table?

onlyjackfrost (commented):
Besides some questions, others LGTM.

.circleci/config.yml
```diff
@@ -137,7 +137,10 @@ jobs:
             - codecov-0.1.0_4653
       - run: rm -rf ./coverage
         # Clear coverage folder
-      - run: yarn nx affected --target=test --ci --coverage --coverageReporters=lcov ${AFFECTED_ARGS}
+      - run:
+          # extend the timeout to 30m to prevent circleci default timeout (10m) reached and stop the long testing.
+          no_output_timeout: 30m
```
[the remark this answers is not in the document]
kokokuo: Thanks for pointing this out, I think we could remove it now because this PR seems not to exceed the timeout when I add new test cases

packages/extension-driver-duckdb/src/lib/duckdbDataSource.ts
kokokuo: The reason is that we will refresh and call the importing method again, but the duckdb could create duplicate index names.

Although we could query whether the index from duckdb and check it exists. But it may do more steps:
1. query the table
2. get the result and do check if exist
3. If not exist, create the index

So, for the simple way, we could drop and recreate directly.

PS: I discovered my `DROP INDEX IF EXISTS ${index} ON ${schema}.${tableName}` syntax is inccoret, the duckdb couldn't drop the index on the specific table. https://duckdb.org/docs/sql/indexes#drop-index

onlyjackfrost (approved):
Thanks for answering, LGTM

packages/extension-driver-duckdb/src/lib/duckdbDataSource.ts
kokokuo: Thanks for @onlyjackfrost feedback the worried cost a lot of time when dropping the index if the table has thousands of thousands ( seems it's an issue in OLTP database, but not sure for OLAP database), we could arrange the performance testing in the future and if it's a issue, we will change back to the above checking steps.

PR #176, opened 2023-05-25
fix(doc): wrong api endpoint

cyyeh:
## Description

In this [page](https://vulcansql.com/docs/get-started/first-api), the api endpoint should be `http://localhost:3000/api/flights` instead of `http://localhost:3000/api/user`

## Issue ticket number

None

## Additional Context

None


wwwy3y3:
Yeah, thanks for correcting it.

wwwy3y3 (approved):

Issue #177, opened 2023-05-26
Support Canner Enterprise's PAT authentication in VulcanSQL 


onlyjackfrost:
## What’s the problem you're trying to solve
Support Canner Enterprise's PAT Authentication method

## Describe the solution you’d like
Add a extension authenticator "CannerPATAuthenticator" to restrict only users of Canner Enterprise could access VulcanSQL API

YAML example:
``` yaml
extensions:
  authenticator-canner: @vulcan/module_path

auth:
  enabled: true

authenticator-canner:
  options:
    canner-pat:
        host: string
        port: number
        ssl: boolean
```

The CannerPATAuthenticator will recognize whether a token is a PAT token with the prefix "Canner-PAT"(case-insensitive) in the authorization header / payload / query string
eg: 
```
Authorization: "Canner-PAT your_canner_pat_token"
```

PR #160, opened 2023-04-13
Feature: Support Catalog User Interface

andreashimin:
## Description

Support the VulcanSQL Catalog user interface for retrieving data in a more efficient and user-friendly manner.

### Sample `vulcan.yaml`
Please notice [`auth`](https://vulcansql.com/docs/api-building/access-control/authenticator) options should be either `basic` or `password-file` is setting up or disabled the auth

```yaml
name: playground1
description: A starter Vulcan project
version: 0.1.0-alpha.1
template:
  provider: LocalFile
  # Path to .sql files
  folderPath: sqls
  codeLoader: InMemory
artifact:
  provider: LocalFile
  serializer: JSON
  # Path to build result
  filePath: result.json
schema-parser:
  reader: LocalFile
  # Path to .yaml files
  folderPath: sqls
document-generator:
  specs:
    - oas3
types:
  - RESTFUL
extensions:
  duckdb: '@vulcan-sql/extension-driver-duckdb'
profiles:
  - profile.yaml
rate-limit:
  options:
    interval:
      min: 1
    max: 10000
enforce-https:
  enabled: false
auth-source:
  options:
    key: token
auth:
  enabled: true
  options:
    basic:
      htpasswd-file:
        path: ./passwd.txt
        users:
          - name: jax
            attr:
              department: engineer
          - name: shimin
            attr:
              department: engineer
    password-file:
      path: ./bcrypt-passwd.txt
      users:
        - name: jax
          attr:
            department: engineer
        - name: shimin
          attr:
            department: engineer
response-format:
  enabled: true
  options:
    default: json
    formats:
      - json
      - csv

```
#### Run Develop Command
```
yarn nx serve catalog-server
```
open `http://localhost:4200`

Default catalog option in yaml
```yaml
catalog:
   hostname: localhost
   port: 4200
```

#### UI
- Catalog list
image

- Catalog detail & preview dataset result
image

- Connect with third-party services tutorial
image

#### Run Develop Command
In `labs/playground1`, run
```
make start
// or
vulcan start // if already build
```
open `http://localhost:3000`

1. Add `/auth/available-types` for retrieving auth type by API
2. Add `catalog-routers` for providing APIs to `Catalog Server`.
    - `/catalog/schemas` 
    - `/catalog/schemas/:sourceName`
The schema APIs above will additionally provide `url: string`, `apiDocUrl: string`, `shareKey: string`
     * `url`: host + urlPath in VulcanSQL
     * `apiDocUrl`: the API document url in VulcanSQL
[...]

PR #181, opened 2023-06-05
Feature: Canner PAT authenticator

onlyjackfrost:
## Description

In this pull request, I implemented an extension "extension-authenticator-canner" that enabled Vulcan API to authenticate users with Canner Enterprise and retrieve the user information from Canner Enterprise. We can access these user attributes after authentication.

Please see the [README](https://github.com/Canner/vulcan-sql/tree/feature/authenticator/packages/extension-authenticator-canner) to know how to setup the extension.

## Issue ticket number

#177 

## Note
We changed the auth configure checking logic to if "auth.enabled" is set to true but didn't configure any authenticator, we will not throw a configured error when serving but throw "all types of authenticator failed" when sending Data API(with or without "Authorization" header or any replacement field/header).

## Test result:

labs/playground
only pat 
image

only basic auth
image

auth is enabled but not configured
image


packages/serve/test/auth/cannerPATAuthenticator.spec.ts
```diff
@@ -0,0 +1,237 @@
+import * as sinon from 'ts-sinon';
+import { IncomingHttpHeaders } from 'http';
+import { Request, BaseResponse } from 'koa';
+import { CannerPATAuthenticator } from '@vulcan-sql/serve/auth';
+import { AuthResult, AuthStatus, KoaContext } from '@vulcan-sql/serve/models';
+
+const wrappedAuthCredential = async (
```
kokokuo, line 7: Wrap the function to reuse it great!

However, according to your test cases for asserting the error,  you should prevent including other non-target test logistics, like here you create the authenticator and called the activate method, it is not the target act code section even though these create heavier and activate behavior seems not possibly cause others errors.

packages/serve/test/auth/cannerPATAuthenticator.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+  ctx: KoaContext,
+  options: any,
+  resolveValue: any = null
+): Promise<AuthResult> => {
+  const authenticator = new CannerPATAuthenticator({ options }, '');
+  if (resolveValue) {
+    sinon.default
+      .stub(authenticator, <any>'fetchCannerUser')
+      .resolves(resolveValue);
```
kokokuo, line 16: Same suggestion as above using `prototype` way.

packages/serve/test/auth/cannerPATAuthenticator.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+  return await authenticator.authCredential(ctx);
+};
+
+const getTokenInfo = async (
+  ctx: KoaContext,
+  options: any
+): Promise<Record<string, any>> => {
+  const authenticator = new CannerPATAuthenticator({ options }, '');
+  await authenticator.activate();
+  return await authenticator.getTokenInfo(ctx);
```
kokokuo, line 28: Same suggestion as above creating the authenticator and calling activate method with the target method at the same place.

packages/serve/test/auth/cannerPATAuthenticator.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+  message,
+});
+const mockOptions = {
+  'canner-pat': {
+    host: 'mockHost',
+    port: 3000,
+    ssl: false,
+  },
+};
+const invalidToken = Buffer.from('clientId:clientSecret').toString('base64');
```
kokokuo, line 49: great way to create the invalid token and also make members know the token content pattern.

kokokuo (commented):
Beside some suggestions, others LGTM 👍

packages/integration-testing/src/examples/example5-get-user-profile-CannerPAT.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+      createdAt: '2023-03-27T12:48:15.882Z',
+      email: 'Alvina_Farrell82@yahoo.com',
+      groups: [{ id: 1, name: 'group1' }],
+    },
+  };
+  // stub fetchCannerUser method in class CannerPATAuthenticator using sinon
+  const stubFetchCannerUser = (user: any) => {
+    const stub = sinon.default.stub(
+      CannerPATAuthenticator.prototype,
+      <any>'fetchCannerUser'
```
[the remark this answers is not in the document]
onlyjackfrost: Thanks for the solution! 
I'll add some comments for now, and I think we can encapsulate how to make requests to Canner and move it to `utils`

packages/serve/src/lib/auth/cannerPATAuthenticator.ts
```diff
@@ -0,0 +1,129 @@
...
+        name: cannerUser.username,
+        attr: restAttrs,
+      },
+    } as AuthResult;
+  }
+
+  private async fetchCannerUser(token: string) {
+    const graphqlUrl = this.getCannerUrl('/web/graphql');
+    try {
+      return await axios.default.post(
```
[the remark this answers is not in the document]
onlyjackfrost: I use `import * as axios from 'axios'`, I'll change the import and use axios.post directly

packages/serve/src/lib/auth/cannerPATAuthenticator.ts
```diff
@@ -0,0 +1,129 @@
...
+        {
+          headers: {
+            Authorization: `Token ${token}`,
+          },
+        }
+      );
+    } catch (error) {
+      throw new InternalError(
+        `Failed to fetch user info from canner server: ${error}`
+      );
```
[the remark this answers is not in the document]
onlyjackfrost: I'll modify this part, and handle the error the same as #172 does

packages/serve/src/lib/auth/cannerPATAuthenticator.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+  private getCannerUrl = (path = '/') => {
+    const { host, port, ssl = false } = this.options;
+    if (process.env['IS_IN_K8S'])
```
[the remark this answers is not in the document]
onlyjackfrost: I'll align it with the prior PR.

packages/integration-testing/src/examples/example2-auth-user.spec.ts
```diff
@@ -2,10 +2,17 @@ import { IBuildOptions, VulcanBuilder } from '@vulcan-sql/build';
...
 import defaultConfig from './projectConfig';
+import faker from '@faker-js/faker';
 
 let server: VulcanServer;
 
+// clear stub after each test
+afterEach(() => {
+  sinon.default.restore();
+});
+
```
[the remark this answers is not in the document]
onlyjackfrost: I'll remove it from the test file.

packages/serve/test/auth/cannerPATAuthenticator.spec.ts
onlyjackfrost: I've removed this function because it is not used after refactoring.

packages/serve/test/auth/cannerPATAuthenticator.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+      status: 500,
+      statusText: 'string',
+      headers: {},
+      config: {},
+    } as any,
+    isAxiosError: true,
+    toJSON: () => ({}),
+    message: 'An error occurred!',
+    name: 'Error',
+  });
```
kokokuo, line 97: Why do you need to mock the `axios.post` reject value ? You could mock the rejected value for `fetchCannerUser` directly.

packages/serve/test/auth/cannerPATAuthenticator.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+    set: sinon.stubInterface<BaseResponse>().set,
+  };
+  const mockResolveValue = {
+    status: 401,
+    data: { error: 'invalid token from canner' },
+  };
+  const authenticator = await getStubAuthenticator(
+    mockOptions,
+    mockResolveValue
+  );
```
kokokuo, line 172: Just curious ? Why do  you pass the `mockResolveValue` be a resolved value for stub `fetchCannerUser` ? 

Seems the `fetchCannerUser` won't return the data whose status is `401`, because when Axios get the `401,` or I miss some situation from canner enterprise ?

kokokuo (commented):
Besides some questions and suggestions, others LGTM 👍  

Btw, we should move our canner PAT authenticator to a new extension package like canner driver, because canner integration is an optional (enhancement) feature for VulcanSQL.

Keep going 👍

kokokuo (commented):
I missed some questions, so I append two in here

packages/serve/test/auth/cannerPATAuthenticator.spec.ts
kokokuo: After discussion, in actually, @onlyjackfrost would like to test the `fetchCannerUser`'s catch and transform part, but currently `fetchCannerUser` is a private method that is called by the private `validate` method, so that why he needs to mock the `axio.post`, so our conclusion is to update the description fo the test case to make others know the intention.

packages/serve/src/lib/auth/cannerPATAuthenticator.ts
```diff
@@ -0,0 +1,129 @@
...
+        message: (err as Error).message,
+      };
+    }
+  }
+
+  private async validate(token: string) {
+    const res = await this.fetchCannerUser(token);
+    if (res.status == 401) throw new UserError('invalid token');
+    if (!res.data.data?.userMe) {
+      throw new InternalError('Can not retrieve user info from canner server');
```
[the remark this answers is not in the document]
onlyjackfrost: I'll remove this part case I've handled the error axios throw that the response status is not 2xx in fetchCannerUser.

packages/serve/src/lib/auth/cannerPATAuthenticator.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    const cannerUser = res.data.data?.userMe;
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
onlyjackfrost: I'll remove the Eslint comment here

packages/serve/src/lib/auth/cannerPATAuthenticator.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+  private getCannerUrl = (path = '/') => {
+    const { host, port, ssl = false } = this.options;
+    if (process.env['IS_ON_KUBERNETES'])
+      return `http://${process.env['WEB_SERVICE_HOST']}${path}`; // for internal usage, we don't need to specify port
+    else {
+      const protocol = ssl ? 'https' : 'http';
+      return `${protocol}://${host}:${port}${path}`;
+    }
```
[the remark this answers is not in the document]
onlyjackfrost: Thanks for the advice! I'll take a look at the document.

packages/serve/test/auth/cannerPATAuthenticator.spec.ts
onlyjackfrost: I'll remove the test cases that are testing the remote response whose status is not 200, thanks

packages/serve/src/lib/auth/cannerPATAuthenticator.ts
```diff
@@ -0,0 +1,129 @@
...
+  private async fetchCannerUser(token: string) {
+    const graphqlUrl = this.getCannerUrl('/web/graphql');
+    try {
+      return await axios.post(
+        graphqlUrl,
+        {
+          operationName: 'UserMe',
+          variables: {},
+          query:
+            'query UserMe{userMe {accountRole attributes createdAt email groups {id name} lastName firstName username',
```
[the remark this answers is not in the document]
onlyjackfrost: These attributes were picked after discussing with William.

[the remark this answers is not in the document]
kokokuo: Thanks for checking the `attribute` field  and that us know it !👍

packages/extension-authenticator-canner/README.md
```diff
@@ -0,0 +1,46 @@
...
+          ssl: true
+   ```
+
+## Testing
+
+```bash
+nx test extension-authenticator-canner
+```
+
+This library was generated with [Nx](https://nx.dev).
```
kokokuo, line 46: Actually, you don't need to add the testing section, because the README will wrap to NPM and publish for user reading, so user will install the NPM and won't need the information like other extension packages.

For the testing information, we will provide it in the development documentation in the future.

labs/playground1/examples/use-canner-pat-authenticator/vulcan.yaml
kokokuo, line 1: If we create a folder and only provide the `vulcan.yaml` sample and not contain other SQL, profiles, or API schema files, then your README of the `extension-authenticator-canner` has done it, so maybe you don't need to add it again.

packages/extension-authenticator-canner/src/test/cannerPATAuthenticator.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  // Act
+  const result = await authenticator.authCredential(ctx);
+
+  // Assert
+  expect(result).toEqual(expected);
+});
+
+// Token info
+it('Should throw when use getTokenInfo with cannerPATAuthenticator', async () => {
```
kokokuo, line 202: Do you miss the `error` word => `Should throw error... ` ?

kokokuo (commented):
Besides 3 suggestions, others LGTM 👍 

Btw, here are 3 PR descriptions for you :)

1. Suggest to add the introduction of how to use Canner PAT extension and its configration in PR descriotion, because it is a new feature, like other PR: https://github.com/Canner/vulcan-sql/pull/160, https://github.com/Canner/vulcan-sql/pull/140, https://github.com/Canner/vulcan-sql/pull/142, https://github.com/Canner/vulcan-sql/pull/108

2. Your current NOTE seems to explain the reason we have the `options` issue but does not mention you solve it, so according to your PR description and title, not know you solve the issue.

3. Suggest you could add another screenshot is that not type `options`, or other users look your PR may consider we must add `options` 


```yaml

auth:
  enabled: true
  # looks like the options also could not add it, right ?
  options:
``

kokokuo (approved):
Thanks for fixing it, Well done, LGTM 👍

PR #182, opened 2023-06-06
Fix: Handling when URL paths are the same but their SQL content is different

### LINKED ISSUE #178: Handling when URL paths are the same but their SQL content is different
## Environment

- Version:
  

  ```
  cli version: 0.4.0
  core version: -
  build version: -
  serve version: -
  ```

- OS: macOS

## What's wrong?
 1. I can see the duplicate API list with the wrong Column Information list on the Vulcan Catalog UI. (Refer to Screenshot UI - 1)
 2. API documentation is inconsistent with Catalog UI information (Refer to Screenshot UI - 2)
 3. Got UserError (Refer to Screenshot UI - 3)

#### Catalog list UI:
截圖 2023-05-29 下午1 30 13

#### Catalog detail page UI (Two `artist/:id` APIs are the same detail page.)
截圖 2023-05-29 下午1 30 24

#### API doc UI
截圖 2023-05-29 下午1 31 09

#### When click a `View API Docs` will see the error
截圖 2023-05-29 下午1 48 05

## Reproducing steps
Step 1: Clone vulcan project
Step 2: go to the `labs/playground1` folder
Step 3: Copy the same file as `labs/playground1/sqls/artist/artist.yaml`, named `artistcopy.yaml`.
Step 4: Copy the same file as `labs/playground1/sqls/artist/artist.sql`, named `artistcopy.sql`. Change its content to
```sql
select ConstituentID, DisplayName, ArtistBio from "artists" where  ConstituentID = {{ context.params.id }}
```
Step 5: Update package.json - add **@vulcan-sql/catalog-server**
```json
{
  "name": "my-first-vulcan-project",
  "dependencies": {
    "@vulcan-sql/catalog-server": "../../dist/packages/catalog-server",
    "@vulcan-sql/core": "0.2.0-dev.20220922.0",
    "@vulcan-sql/extension-driver-duckdb": "0.2.0-dev.20220919.0",
    "@vulcan-sql/serve": "0.2.0-dev.20220922.0"
  },
  "devDependencies": {
    "@vulcan-sql/build": "0.2.0-dev.20220922.0"
  }
}

```
Step 6: Enable auth for `vulcan.yaml`
Step 7: execute `make` command
Step 8: execute vulcan start and vulcan catalog commands in different terminal windows
Step 9: open http://localhost:4200/catalog

### Screenshot of Editor
截圖 2023-05-29 下午1 37 06

kokokuo:
## Description
From #178, we discover the issue happened because "urlPath" could be duplicated in different schemas, then when API was created, users could request, but only one SQL file could be executed, and Catalog UI could see the duplicate API URL path, however, API request should be unique and have different query behavior.

## Solution
The correct behavior is that API urlPath is unique, so we do not allow duplicate "urlPath" value in the schemas, therefore, we will add the validation to check all schemas before vulcanSQL build the templates and schemas to the artifact.

Otherwise, we also need to validate the same duplicate "urlPath" value in the schemas after the artifact loads to VulcanSQL, but not yet to mount APIs and start the server in VulcanSQL, the reason is that our VulcanSQL could also allow the user set the artifact directly in `vulcan.yaml` and use the artifact to run VulcanSQL by `vulcan serve`, so if the artifact has duplicated "urlPath" we should also check it. 

## Issue ticket number

closes #178

## Additional Context
We do not add the validation in the artifact builder to validate it, because if we check when artifactBuilder calling `build`, it has including the unnecessary generate specification steps. otherwise, we hope the artifact builder could focus on build and load behavior.


fredalai (approved):
LGTM, thanks for fixing the bug

PR #184, opened 2023-06-08
Chore: Add additional catalog package.json

andreashimin:
## Description

Nx will not help to create `package.json` & `README.md` in the source of the `Next.js` package by default,
so the Catalog server needs to add `package.json` to provide `name` & `version` and add `README.md` for release.

## Additional Context
- remove default project setting

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -27,6 +27,5 @@
         "linter": "eslint"
       }
     }
-  },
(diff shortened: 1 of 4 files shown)


kokokuo (approved):
Thanks for making the catalog-server could generate the correct name package and also provide the document, LGTM 👍

PR #196, opened 2023-06-26
Fix(extension-canner-authenticator): fix endpoint to canner web service

onlyjackfrost:
## Description

This PR fixes the endpoint to web service, the endpoint is different between connecting from external and internal. 

## Additional Context

When connecting from the external network, the nginx will route the requests to web service when the request starts with '/web', and rewrite the request.

## Test

can connect to web service from outside the k8s 
image

### DIFF
--- a/packages/extension-authenticator-canner/src/lib/authenticator/pat.ts
+++ b/packages/extension-authenticator-canner/src/lib/authenticator/pat.ts
@@ -81,7 +81,7 @@ export class CannerPATAuthenticator extends BaseAuthenticator<CannerPATOptions> 
   }
 
   private async fetchCannerUser(token: string) {
-    const graphqlUrl = this.getCannerUrl('/web/graphql');
+    const graphqlUrl = this.getCannerUrl('/graphql');
     try {
       return await axios.post(
         graphqlUrl,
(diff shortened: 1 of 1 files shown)


packages/extension-authenticator-canner/src/lib/authenticator/pat.ts
```diff
@@ -110,11 +110,14 @@ export class CannerPATAuthenticator extends BaseAuthenticator<CannerPATOptions>
...
   private getCannerUrl(path = '/') {
     const { host, port, ssl = false } = this.options;
-    if (config.isOnKubernetes)
-      return `http://${process.env['WEB_SERVICE_HOST']}${path}`; // for internal usage, we don't need to specify port
+    // the web endpoint is different between connecting from internal or external
+    if (config.isOnKubernetes) return `${process.env['WEB_ENDPOINT']}${path}`;
     else {
       const protocol = ssl ? 'https' : 'http';
-      return `${protocol}://${host}${port ? `:${port}` : ''}${path}`;
+      // our nginx proxy will redirect all request to web service when path starts with /web
```
kokokuo, line 117: I think the comment could be put in the above condition and simply modify like the below:
```ts
On K8s, we use nginx proxy and it will redirect all requests to web service when the path starts with /web, so we don't need to add the path prefix `/web`.
```
because nginx redirects to `/web` related to the remote environment on k8s.

kokokuo (commented):
Simple suggestion for comment, others LGTM

kokokuo (approved):
LGTM 👍

PR #201, opened 2023-06-29
Chore(authenticator): match authenticator in middleware

onlyjackfrost:
## Description
Refactor the authentication middleware to match the authenticator first then validate the user with the matched authenticator 

## Additional Context
We still having "INDETERMINATE" state in the build-in authenticators and check the authenticator is set proper, because build-in authenticators will always be initialized and the scenario is user send a request with the token identifier which is the authenticator that did not been configured in the vulcan.yaml file


onlyjackfrost:
close this PR, because branch name

PR #203, opened 2023-06-29
Chore(Auth Middleware): match authenticator in middleware

onlyjackfrost:
## Description
Refactor the authentication middleware to match the authenticator first then validate the user with the matched authenticator 

## Additional Context
We still having "INDETERMINATE" state in the build-in authenticators and check the authenticator is set proper, because build-in authenticators will always be initialized and the scenario is user send a request with the token identifier which is the authenticator that did not been configured in the vulcan.yaml file

### DIFF
--- a/packages/serve/test/auth/simpleTokenAuthenticator.spec.ts
+++ b/packages/serve/test/auth/simpleTokenAuthenticator.spec.ts
@@ -55,7 +55,7 @@ describe('Test simple-token authenticator', () => {
     },
   ] as SimpleTokenOptions;
 
-  it.each([[{}]])(
+  it.each([[{}], [{ 'simple-token': [] }]])(
     'Should auth incorrect when options = %p in options',
     async (options) => {
       // Arrange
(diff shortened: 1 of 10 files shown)


kokokuo (commented):
Besides some questions,s other LGTM 👍  Thanks for updating the test cases!

packages/serve/test/auth/simpleTokenAuthenticator.spec.ts
```diff
@@ -55,7 +55,7 @@ describe('Test simple-token authenticator', () => {
     },
   ] as SimpleTokenOptions;
 
-  it.each([[{}], [{ basic: {} }], [{ 'simple-token': [] }]])(
+  it.each([[{}]])(
```
kokokuo, line 58: same as above

kokokuo (commented):
add one same question as above to point it.

kokokuo (approved):
LGTM 👍

PR #217, opened 2023-07-05
add examples

cyyeh:
## Description

Add examples folder which showcases how VulcanSQL can be used

## Issue ticket number

None

## Additional Context

The first example is a COVID-19 Global Data Visualizations using VulcanSQL and Streamlit.


andreashimin:
Hi @cyyeh @wwwy3y3 
Would it be more appropriate if we could put examples in other repo like https://github.com/Canner/vulcan-demo or new repo like vulcan-sql-examples?
Since this example not really related with the source code.

wwwy3y3:
Hmm, alright. @cyyeh let's move to another repo.

Issue #231, opened 2023-07-11
Support reading data from data lakes.


onlyjackfrost:
## What’s the problem you're trying to solve
The problem I am trying to solve is integrating with data lakes provided by AWS and GCP. The datasets might be stored in these data lakes, so it is important to consider integrating with them effectively.

## Describe the solution you’d like
DuckDB provides an extension called "httpfs" that enables reading from and writing to remote files. This feature can be utilized to integrate with data lakes such as AWS and GCP.

PR #234, opened 2023-07-12
feature: support reading data from data lake

onlyjackfrost:
## Description

This PR uses "[httpfs](https://duckdb.org/docs/extensions/httpfs)" extension and  allows DuckDB to read remote data.
The remote files which were partitioned using "Hive Partition" is also supported. Read the [doc](https://duckdb.org/docs/data/partitioning/hive_partitioning#hive-partitioning) for more.

## Issue ticket number

#231 

## Additional Context
The configuration parameters set in the profile.yaml file were used as the default credential when sending request to DuckDB,  To read data using a different credential, utilize the [Per-Request Configuration](https://duckdb.org/docs/extensions/httpfs#per-request-configuration) and concat the credential after your object url

Example: 
```
# example from DuckDB doc
SELECT *
FROM 's3://bucket/file.parquet?s3_region=region&s3_session_token=session_token' T1
INNER JOIN 's3://bucket/file.csv?s3_access_key_id=accessKey&s3_secret_access_key=secretKey' T2;
```

## Test Result
Use playground to test reading hive partitioned .parquet file from s3 
image

### DIFF
--- a/packages/extension-driver-duckdb/src/lib/duckdbDataSource.ts
+++ b/packages/extension-driver-duckdb/src/lib/duckdbDataSource.ts
@@ -15,6 +15,7 @@ import {
 } from '@vulcan-sql/core';
 import * as path from 'path';
 import { buildSQL } from './sqlBuilder';
+import { DuckDBExtensionLoader } from './duckdbExtensionLoader';
 
 const getType = (value: any) => {
   const jsType = typeof value;
@@ -29,7 +30,7 @@ const getType = (value: any) => {
 };
 
 // read more configuration in DuckDB document: https://duckdb.org/docs/extensions/httpfs
-export interface S3Credentials {
+export interface ConfigurationParameters {
   region?: string;
   accessKeyId?: string;
   secretAccessKey?: string;
@@ -44,7 +45,7 @@ export interface DuckDBOptions {
   'persistent-path'?: string;
   'log-queries'?: boolean;
   'log-parameters'?: boolean;
-  's3-credentials'?: S3Credentials;
+  'configuration-parameters'?: ConfigurationParameters;
 }
 
 @VulcanExtensionId('duckdb')
@@ -56,7 +57,7 @@ export class DuckDBDataSource extends DataSource<any, DuckDBOptions> {
       db: duckdb.Database;
       logQueries: boolean;
       logParameters: boolean;
-      s3Credentials: S3Credentials;
+      configurationParameters: ConfigurationParameters;
     }
   >();
   private logger = this.getLogger();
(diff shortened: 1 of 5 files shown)


packages/extension-driver-duckdb/src/lib/duckdbDataSource.ts
```diff
@@ -28,18 +29,36 @@ const getType = (value: any) => {
...
+  endpoint?: string;
+  url_style?: string;
+  use_ssl?: boolean;
+}
+
 export interface DuckDBOptions {
   'persistent-path'?: string;
   'log-queries'?: boolean;
   'log-parameters'?: boolean;
+  'configuration-parameters'?: ConfigurationParameters;
```
andreashimin, line 48: Is it need to write down in `README.md` for users?

packages/extension-driver-duckdb/src/lib/duckdbExtensionLoader.ts
```diff
@@ -0,0 +1,74 @@
...
+    try {
+      conn.run(`LOAD ${extensionName}`);
+    } catch (error) {
+      this.logger.debug(`Error when loading extension:${extensionName}`);
+      throw error;
+    }
+    try {
+      await Promise.all(
+        Object.entries(extensionConfigurations).map(
+          async ([dbParameterName, configurationKey]) => {
```
andreashimin, line 48: Why here need to use `async` if it's no `await` inside the map function?

andreashimin (approved):
Besides some questions,
Others LGTM 👍

packages/extension-driver-duckdb/src/lib/duckdbExtensionLoader.ts
```diff
@@ -0,0 +1,74 @@
...
+export enum HTTPFS_CONFIGURATIONS {
+  s3_region = 'region',
+  s3_access_key_id = 'accessKeyId',
+  s3_secret_access_key = 'secretAccessKey',
+  // sessionToken= alternative option for accessKeyId and secretAccessKey
+  s3_session_token = 'sessionToken',
+  s3_endpoint = 'endpoint',
+  s3_url_style = 'url_style',
+  s3_use_ssl = 'use_ssl',
+}
```
andreashimin, line 15: Another suggestion~
I will recommend not using `enum` for configurations because it is not designed for the purpose
it could change to simply use an `object` to map instead, or also can do `freeze` to prevent value change if needed:

```ts
const HTTPFS_CONFIGURATIONS = Object.freeze({
  s3_region: 'region',
  s3_access_key_id: 'accessKeyId',
  ...
})
```

kokokuo (approved):
Besides some suggestions, Others LGTM 👍  You could decide to refactor it or not.

packages/extension-driver-duckdb/src/lib/duckdbDataSource.ts
```diff
@@ -221,4 +245,34 @@ export class DuckDBDataSource extends DataSource<any, DuckDBOptions> {
...
   }
+
+  private async initDatabase(dbPath: string) {
+    const db = new duckdb.Database(dbPath);
+    const conn = db.connect();
+    await this.installExtensions(conn);
+    return db;
+  }
+
+  private async installExtensions(
```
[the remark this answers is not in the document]
onlyjackfrost: I was thinking about this last night.

PR #242, opened 2023-07-20
Chore: update release assets info of the binary process

fredalai:
## Description

In the current Release process, we will compress the constructed binary files and put them into the assets in the release notes. The file format is `vulcan.{os}.{version}.{tar.gz/zip}`.

We should simplify. As we know, each Release Note not only has the assets part but also has the Release version number, so we will change the file name to no longer contain the version number part.

So, we will have the following changes:
1. Change the file name format.
   - Ubuntu: **vulcan.linux.tar.gz**
   - Mac OS: **vulcan.osx.tar.gz**
   - Windows: **vulcan.win.zip**

2. It is no longer necessary to change the download link in the docs

## Issue ticket number

closes #xx

## Additional Context

Reference doc: https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases#linking-to-the-latest-release

### DIFF
--- a/packages/doc/docs/get-started/installation.mdx
+++ b/packages/doc/docs/get-started/installation.mdx
@@ -11,22 +11,22 @@ We provide pre-built binaries for Linux, macOS, and Windows. The binary includes
 <Tabs groupId="install-os">
 <TabItem value="osx" label="MacOS">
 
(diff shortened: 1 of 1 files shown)


kokokuo (approved):
LGTM 👍

PR #256, opened 2023-07-27
Fix: Rename the downloaded parquet file to add parquet extension

kokokuo:
## Description
When using  `exetnsion-driver-canner` and caching dataset feature, the parquet file was downloaded from Canner Enterprise, but not find it in the local folder and cause vulcan-sql to fail.
MicrosoftTeams-image (1)

MicrosoftTeams-image (2)

The reason is that the downloaded parquet file does not include `parquet` extension name, so it caused the cache layer could not to find the parquet file and failed. After renaming the filename when downloading, the issue has been fixed, see:

截圖 2023-07-27 下午2 18 30

截圖 2023-07-27 下午4 53 33

## Issue ticket number

closes #xx

### DIFF
--- a/labs/playground1/Makefile
+++ b/labs/playground1/Makefile
@@ -6,7 +6,7 @@ start: build test-data/moma.db ../../node_modules
 	@vulcan start
 
 # build the required packages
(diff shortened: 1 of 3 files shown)


onlyjackfrost (approved):
LGTM

PR #262, opened 2023-07-31
Fix: assign canner driver connection pool by env

onlyjackfrost:
## Description

assign canner driver connection pool max limit by env


kokokuo (approved):
LGTM 👍

PR #268, opened 2023-08-03
fix issue#261 and add bq to playground1

### LINKED ISSUE #261: Couldn't use caching when the connector is BigQuery
## Environment

- Version:
  

  ```
  cli version: 0.7.1
  core version: 0.7.1
  build version: 0.7.1
  serve version: 0.7.1
  ```

- OS: macOS with M1 chip

## What's wrong?

Couldn't use caching when connector is BigQuery. After discussing with @onlyjackfrost, there might be two places we need to modify in `packages/extension-driver-bq/src/lib/bgDataSource.ts`:
```typescript
// Use the directory to avoid filename collision,
    // The wildcard indicates the partition,
    // ref: https://cloud.google.com/bigquery/docs/reference/standard-sql/other-statements#export_option_list
    const bucketObjPrefix = path.join(bucketName, directory);
    const uri = `gs://${path.join(bucketObjPrefix, 'part*.parquet')}`;
    const queryOptions = {
      query: `EXPORT DATA OPTIONS( uri="${uri}", format='PARQUET') AS ${statement}`,
      location: options?.location || 'US',
    };
```

```
const getFilesResponse = await storage.bucket(bucketName).getFiles({
  prefix: directory.slice(1),
})
```

Users can caching when the connector is BigQuery

cyyeh:
## Description

- Fix the issue that we can't export BigQuery data to GCS and download parquet files from GCS.
 -  Root cause: URI issue, originally we gave wrong paths to BigQuery and GCS
- Add BigQuery data source in playground1
- Update package.json dependencies in playground1

## Issue ticket number

closes #261

## Additional Context

To double check in playground1:
1. Change your gcp service account key file path in profile.yaml
2. In playground1, run `yarn`
3. In playground1, run `make`

If it's successful, you'll see a new parquet file downloaded into a directory with name sth like this `tmp/tpch/bq/tpch_orders/xxxx/partxxxx.parquet`

### DIFF
--- a/labs/playground1/.gitignore
+++ b/labs/playground1/.gitignore
@@ -3,3 +3,4 @@ result.json
 spec-oas3.yaml
 .vulcan-debug
 dist
+tmp
--- a/labs/playground1/Makefile
+++ b/labs/playground1/Makefile
@@ -6,7 +6,10 @@ start: build test-data/moma.db ../../node_modules
 	@vulcan start
 
 # build the required packages
-build: pkg-core pkg-build pkg-serve pkg-catalog-server pkg-cli pkg-extension-driver-duckdb pkg-extension-authenticator-canner pkg-extension-driver-canner pkg-extension-driver-clickhouse pkg-extension-huggingface
+build: pkg-core pkg-build pkg-serve pkg-catalog-server pkg-cli pkg-extension-driver-duckdb \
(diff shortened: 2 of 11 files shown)


kokokuo (approved):
LGTM 👍

PR #273, opened 2023-08-06
fix wrong api urls on openapi doc

cyyeh:
## Description

Fix wrong api URL path on Open API doc. Since the url path of API generated by VulcanSQL will start from '/api', but there's no `/api` prefix shown on Open API doc.

## Issue ticket number

None


onlyjackfrost (approved):
LGTM

PR #277, opened 2023-08-13
add extension-api-caller

cyyeh:
## Description

Add extension-api-caller that allows users to call APIs to get data from the Internet. As of now, it only supports REST APIs.

## Issue ticket number

None


packages/extension-api-caller/src/lib/filters/restApiCaller.ts
```diff
@@ -0,0 +1,37 @@
...
+    params: args['arg'] && args['arg'] !== ':id' ? { [args['arg']]: value } : {},
+  }
+  if (args['body']) {
+    options = {...options, body: JSON.parse(args['body'])}
+  }
+  if (args['headers']) {
+    options = {...options, headers: JSON.parse(args['headers'])}
+  }
+
+  const results = await axios(options);
```
onlyjackfrost, line 29: We should handle non status 2xx response with try cache

packages/extension-api-caller/src/lib/filters/restApiCaller.ts
```diff
@@ -0,0 +1,37 @@
...
+  let options: any = {
+    url: url,
+    method: httpMethod,
+    params: args['arg'] && args['arg'] !== ':id' ? { [args['arg']]: value } : {},
+  }
+  if (args['body']) {
+    options = {...options, body: JSON.parse(args['body'])}
+  }
+  if (args['headers']) {
+    options = {...options, headers: JSON.parse(args['headers'])}
```
kokokuo, line 26: After discussing with @cyyeh and @onlyjackfrost, we decide to design to pass the `header`, `query, `body`, and `path` from the `value` through the pipe operator, and `rest_api` only pass the `url` keyword argument in the filter function, looks like below

If the URL need to pass the path parameter e.g: `http://localhost:3000/users/:id/workspaces/:sqlName` , then the SQL will like:

```sql
{% set value = { "path": { "id": 1, "sqlName": "ws_1"} }
SELECT {{ value | reset_api(url=http://localhost:3000/users/:id/workspaces/:sqlName) }}
```

If the URL is `http://localhost:3000/users` and could query by query string:

```sql
{% set value = { "query": { "name": "andy" } }
SELECT {{ value | reset_api(url=http://localhost:3000/users) }}
```

If the URL is `http://localhost:3000/users` and could pass the value by header:

```sql
{% set value = { "header": { "x-canner-token": "xxxxx" } }
SELECT {{ value | reset_api(url=http://localhost:3000/users) }}
```

And also update the sample in the test cases

If the URL is `http://localhost:3000/users` and supports `POST` method and pass by `body`: 

```sql
{% set value = { "body": { name: "andy" } }
SELECT {{ value | reset_api(url=http://localhost:3000/users, method="POST") }}
```

kokokuo (commented):
Beside some suggestion, others LGTM 👍

kokokuo (commented):
LGTM 👍

kokokuo (approved):

PR #278, opened 2023-08-14
Feature: improve duckdbDataSource performance

onlyjackfrost:
## Description

- Add performance analysis tool
- Modify Duckdb data source execute method: 
    - Pararelize execution and use connection.all  instead of nextChunk

Note: 
With the same data, the performance of stream.nextChunk is 1.5~2 times of the performance of connection.all

### DIFF
--- a/packages/extension-driver-duckdb/src/lib/sqlBuilder.ts
+++ b/packages/extension-driver-duckdb/src/lib/sqlBuilder.ts
@@ -42,7 +42,7 @@ export const buildSQL = (
   builtSQL = addLimit(builtSQL, operations.limit);
   builtSQL = addOffset(builtSQL, operations.offset);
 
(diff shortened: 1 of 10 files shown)


packages/extension-driver-duckdb/src/lib/sqlBuilder.ts
```diff
@@ -26,15 +26,23 @@ export const isNoOP = (
   return true;
 };
 
+const duckdbExecuteChunkSize =
+  process.env['DUCKDB_EXECUTE_CHUNK_SIZE'] || '2000';
```
kokokuo, line 30: Suggest to update to mentioned the env in README, and what influence will be if set the `env` to large value or small value.

kokokuo (commented):
Beside some suggestion, others LGTM

kokokuo (approved):
LGTM  👍  Thanks for fixing.

PR #283, opened 2023-08-18
Fix: caching layer persistent

onlyjackfrost:
## Description

 - Make caching layer datasource can be persistent 
    - It's will reduce the memory usage when loading parquet file to cache 
 - Skip exporting data if parquet file exists when loading cache
    - avoid exporting/downloading data if the parquet files in path is still fresh

### DIFF
--- a/packages/core/src/lib/cache-layer/cacheLayerLoader.ts
+++ b/packages/core/src/lib/cache-layer/cacheLayerLoader.ts
@@ -43,31 +43,40 @@ export class CacheLayerLoader implements ICacheLayerLoader {
     templateName: string,
     cache: CacheLayerInfo
   ): Promise<void> {
-    const { cacheTableName, sql, profile, indexes } = cache;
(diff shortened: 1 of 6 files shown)


packages/core/src/lib/cache-layer/cacheLayerLoader.ts
```diff
@@ -81,4 +90,16 @@ export class CacheLayerLoader implements ICacheLayerLoader {
...
+
+  private getParquetFiles(directory: string): string[] {
+    if (!directory || !fs.existsSync(directory)) return [];
+    const files = fs.readdirSync(directory);
+    const parquetFiles = [];
+    for (const file of files) {
+      if (/\.parquet$/.test(file)) {
+        parquetFiles.push(file);
+      }
+    }
```
andreashimin, line 102: (NIT) Prefer functional programming for filtering
```ts
const parquetFiles = files.filter(file => /\.parquet$/.test(file))
```

andreashimin (approved):
LGTM

PR #286, opened 2023-08-21
Chore: add setting duckdb config via env

onlyjackfrost:
enable setting duckdb config after open database and before exec queries


kokokuo (approved):
Urgent feature, LGTM 👍 

Please refactor, and update the README, and document in the future.

PR #290, opened 2023-08-29
Fix: Evict unused parquet file after exporting cache data

onlyjackfrost:
## Description

Remove the dormant Parquet files to free up hardware storage as these files are no longer required.

## Test
Using playground to test, under the tmp/artist-artist/canner/lineitem folder, the elder parquet files were evict after exporting data successfully.
image


kokokuo (approved):
Perfect ! LGTM 👍

PR #292, opened 2023-08-30
fix #236, can't write comments in sql files

### LINKED ISSUE #236: Can't write comments in sql files
## Environment

- Version:
  

  ```
  cli version: 0.6.0
  core version: 0.6.0
  build version: 0.6.0
  serve version: 0.6.0
  ```

- OS: macOS with M1 chip
- Node.js: v16.20.0

## What's wrong?

Comments do not work in SQL files. After I commented on the code, the API didn't change.

example code snippet 
```sql
-- SELECT 
--     businesses.*,
--     {% masking 'tips.user_id' partial(2, 'xxxxxxx', 2) %} AS tip_user_id,
--     tips.text AS tip_text,
--     tips.date AS tip_date,
--     tips.compliment_count AS tip_compliment_count
-- FROM 'data/yelp_academic_dataset_business.parquet' AS businesses
-- INNER JOIN (
--     SELECT * FROM 'data/yelp_academic_dataset_tip.parquet'
-- ) tips ON (
--     businesses.business_id = tips.business_id
-- )
-- WHERE businesses.business_id = {{ context.params.business_id | is_required }}

SELECT * FROM 'data/yelp_academic_dataset_business.parquet' AS businesses
```

I can write comments(single line and multi line) on SQL files.

## Reproducing steps

Try to add comments in SQL files

cyyeh:
## Description

Remove comments in sql statements

## Issue ticket number

closes #236 

## Additional Context

Previously, we didn't remove comments from sql statements.


packages/core/test/template-engine/built-in-extensions/query-builder/builder.spec.ts
```diff
@@ -167,3 +167,43 @@ it('Extension should throw error when no profile defined', async () => {
     `No profile name found`
   );
 });
+
+it('xtension should remove comments in sql statements', async () => {
```
kokokuo, line 171: Typo: `xtension` should be `Extension`

packages/core/test/template-engine/built-in-extensions/cache/cache.spec.ts
```diff
@@ -174,3 +174,41 @@ select count(*) as count from vulcan.user where user.id = {{ context.params.user
   expect(binding[0].get('$1')).toBe(`user-id`);
   expect(resultData).toEqual([{ count: 1 }]);
 });
+
+it('Extension should remove comments in sql statements', async () => {
```
kokokuo, line 178: Please assist to add the multiple line comment test case, thanks !

packages/core/test/template-engine/built-in-extensions/query-builder/builder.spec.ts
kokokuo: BTW, Please assist to add the multiple line comment test case, thanks !

kokokuo (commented):
Beside multiple test case suggestion and one typo, others LGTM

packages/core/test/template-engine/built-in-extensions/query-builder/builder.spec.ts
cyyeh: fixed

packages/core/test/template-engine/built-in-extensions/cache/cache.spec.ts
cyyeh: added multi-line comments in the test case

kokokuo (approved):
LGTM 👍  Perfect !

oscar60310:
Hey @kokokuo @cyyeh, this PR does resolve the issue with the comments, but there are some trade-offs involved in doing so.

The syntax `--` is the comments for **SQL Engines**, not for Jinja, which uses `{# #}`. So this will work:
```jinja
{# {{ context.params.business_id  }} #}
SELECT * FROM 'data/yelp_academic_dataset_business.parquet' AS businesses
```

Vulcan treats `--` as a normal string so it will try to parametrize it. So this won't work:
```sql
-- {{ context.params.business_id  }}
SELECT * FROM 'data/yelp_academic_dataset_business.parquet' AS businesses
```
```sql
-- $1
SELECT * FROM 'data/yelp_academic_dataset_business.parquet' AS businesses
```
SQL engine complains about that because of number of bindings doesn't match.

If we can prevent the variables in SQL engine comments from being parametrized, the issue will be resolved.

---

There are some cases we might break:

1. If the SQL engine DOES need the comments, e.g. for auditing.
    ```sql
    -- Sent from Vulcan SQL: {{ context.user.name | raw }}
    SELECT * FROM xxxxx
    ```
2. Special comments, e.g. [optimizer hints](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html).
    ```sql
    SELECT /*+ BKA(t1) NO_BKA(t2) */ * FROM t1 INNER JOIN t2 WHERE ...;
    ```
3. Special syntax. e.g.
    ```sql
    SELECT {{ 1--1 }} AS RESULT; {# should be 2 #}
    ```

PR #295, opened 2023-09-04
Feature(Core, Extension-driver-canner): pass headers down to datasource & use info in API headers to connect to canner

onlyjackfrost:
## Description

To use the API request context(e.g.: some customized headers) in dataSource, we need to first inject the API context into NunjucksExecutionMetadata which stores global info and passes around tag runners, and inject API context into dataSource in each tag runner.

## Note
Pass API request context in the Nunjucks filter is not supported yet 

## Test
unit test ✅ 
manual integration test using canner driver ✅ 
 - can use headers to access canner

截圖 2023-09-04 下午2 41 44

### DIFF
--- a/packages/core/src/lib/data-query/builder/dataQueryBuilder.ts
+++ b/packages/core/src/lib/data-query/builder/dataQueryBuilder.ts
@@ -3,6 +3,7 @@ import {
   Pagination,
   DataResult,
   isOffsetPagination,
+  IncomingHttpHeaders,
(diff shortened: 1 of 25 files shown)


kokokuo:
Please assist in making the code coverage to up 80%, You could check the code coverage in the local by `nx affected:test --all --coverage` first to know the reason.

kokokuo (approved):
LGTM 👍  Thanks for fixing !

PR #300, opened 2023-09-11
Feature(CORE): Activity log extension

onlyjackfrost:
## Description

Add a activity log extension to send activity log like incoming request information and cache refreshing result.
Activity logs aim to provide a more specific and accurate information compared to the access log or debug logger stdout. 

## How to use
```
# vulcan.yaml
activity-log:
    enabled: true
    options:
        http-logger:
            connection:
                ssl: boolean. # use https or not
                host: string
                port: integer
                path: "url path"
```
After you set your project config, vulcan server will send activity log to remote server asynchronously when cache refreshing job is done and after processed an API request.

A cache refreshing activity log will contain info
```
logTime(timestamp in UTC) -- the time that refresh job started
urlPath -- indicate which API it belong
sql -- the sql statement that will be executed 
refreshResult -- Enum(success, failed), the result of refresh job
```

In an API request activity log will contain info
```
logTime(timestamp in UTC) -- the timestamp that server received the request
duration(ms) -- the approximate time that server handled the API request
method -- the request method
url -- the request url
ip -- the original IP
header -- the request headers
params -- the request params
query -- the request query parameters
status -- the response HTTP status
error -- the error message that VulcanSQL provided
user -- the attribute that the client user have (gain after authenticate)
```  

## Test
unit tests
Labs:
manually create a koa server on localhost with a router  "(POST)/test"
vulcan.yaml:
```
...
activity-log:
  enabled: true
  options:
    http-logger:  
      connection:
        host: localhost
        port: 8080
        path: test
```
 - Activity log should be send after refresh job done

image

image

 - Activity log should be send after processed API request
image

### DIFF
--- a/packages/core/src/lib/cache-layer/cacheLayerRefresher.ts
+++ b/packages/core/src/lib/cache-layer/cacheLayerRefresher.ts
@@ -100,7 +100,6 @@ export class CacheLayerRefresher implements ICacheLayerRefresher {
   ) {
     const { urlPath } = schema;
     const { sql } = cache;
-    // if fn is not a function, return
     let refreshResult = RefreshResult.SUCCESS;
     const now = moment.utc().format('YYYY-MM-DD HH:mm:ss');
     const templateName = schema.templateSource.replace('/', '_');
--- a/packages/core/src/lib/cache-layer/cacheLayerRefresher.ts
+++ b/packages/core/src/lib/cache-layer/cacheLayerRefresher.ts
@@ -72,14 +72,14 @@ export class CacheLayerRefresher implements ICacheLayerRefresher {
               const refreshJob = new SimpleIntervalJob(
                 { milliseconds, runImmediately },
                 new AsyncTask(workerId, async () => {
-                  await this.sendActivityLogAfterLoad(schema, cache);
+                  await this.loadCacheAndSendActivityLog(schema, cache);
                 }),
                 { preventOverrun: true, id: workerId }
               );
               // add the job to schedule cache refresh task
               this.scheduler.addIntervalJob(refreshJob);
             } else {
-              await this.sendActivityLogAfterLoad(schema, cache);
+              await this.loadCacheAndSendActivityLog(schema, cache);
             }
           })
         );
@@ -94,7 +94,7 @@ export class CacheLayerRefresher implements ICacheLayerRefresher {
     this.scheduler.stop();
   }
 
-  private async sendActivityLogAfterLoad(
+  private async loadCacheAndSendActivityLog(
     schema: APISchema,
     cache: CacheLayerInfo
   ) {
--- a/packages/core/src/lib/cache-layer/cacheLayerRefresher.ts
+++ b/packages/core/src/lib/cache-layer/cacheLayerRefresher.ts
@@ -5,6 +5,8 @@ import { inject, injectable, multiInject } from 'inversify';
 import { TYPES } from '@vulcan-sql/core/types';
 import {
   APISchema,
+  ActivityLogContentOptions,
+  ActivityLogType,
   CacheLayerInfo,
   IActivityLogger,
 } from '@vulcan-sql/core/models';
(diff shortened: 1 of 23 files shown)


packages/core/src/lib/cache-layer/cacheLayerRefresher.ts
```diff
@@ -77,6 +94,45 @@ export class CacheLayerRefresher implements ICacheLayerRefresher {
     this.scheduler.stop();
   }
 
+  private async sendActivityLogAfterLoad(
+    schema: APISchema,
+    cache: CacheLayerInfo
+  ) {
+    const { urlPath } = schema;
+    const { sql } = cache;
+    // if fn is not a function, return
```
kokokuo, line 103: What is the comment? Seems no `fn` ?

packages/core/src/lib/loggers/httpLogger.ts
```diff
@@ -0,0 +1,61 @@
...
+      this.logger.debug(`Activity log sent`);
+    } catch (err) {
+      this.logger.debug(
+        `Failed to send activity log to http logger, url: ${url}`
+      );
+      throw err;
+    }
+  }
+
+  protected sendActivityLog = async (
```
kokokuo, line 45: Why does here use the functional expression repsentation ( `xxx = () => {}` )? Is the reason to do it?

packages/core/src/lib/loggers/httpLogger.ts
```diff
@@ -0,0 +1,61 @@
...
+    url: string,
+    payload: any,
+    headers: AxiosRequestHeaders | undefined
+  ): Promise<void> => {
+    await axios.post(url, payload, {
+      headers: headers,
+    });
+  };
+
+  protected getUrl = (connection: HttpLoggerConnectionConfig): string => {
```
kokokuo, line 55: Same as above.

packages/core/src/lib/loggers/httpLogger.ts
```diff
@@ -0,0 +1,61 @@
...
+  private logger = this.getLogger();
+
+  public async log(payload: any): Promise<void> {
+    if (!this.isEnabled()) return;
+    const option = this.getOptions();
+    if (!option?.connection) {
+      throw new Error('Http logger connection should be provided');
+    }
+    const headers = option.connection.headers;
+    const url = this.getUrl(option.connection);
```
kokokuo, line 32: Just curious seems you have provided a function called `getUrl` in `utils` and here you also create the same name method, but also the same code ( not using the `header` ), so why does not use the `getUrl` in `utils` directly? 

```
const url = getUrl(connection as ConnectionConfig);
```

packages/core/src/lib/utils/url.ts
```diff
@@ -0,0 +1,14 @@
+export interface ConnectionConfig {
+  ssl?: boolean | undefined;
+  host?: string | undefined;
```
kokokuo, line 3: You have added the `?`, is means undefined, so actually, you don't need to write again by `undefined`.

packages/core/src/lib/utils/url.ts
```diff
@@ -0,0 +1,14 @@
...
+  port?: number | string;
+  path?: string | undefined;
+}
+
+export const getUrl = (connection: ConnectionConfig): string => {
```
kokokuo, line 8: Same as above.

kokokuo (commented):
Beside some question and suggestion, other LGTM 👍

packages/core/src/lib/cache-layer/cacheLayerRefresher.ts
```diff
@@ -53,16 +72,14 @@ export class CacheLayerRefresher implements ICacheLayerRefresher {
...
-                  await this.cacheLoader.load(templateName, cache);
+                  await this.sendActivityLogAfterLoad(schema, cache);
                 }),
                 { preventOverrun: true, id: workerId }
               );
               // add the job to schedule cache refresh task
               this.scheduler.addIntervalJob(refreshJob);
             } else {
-              await this.cacheLoader.load(templateName, cache);
+              await this.sendActivityLogAfterLoad(schema, cache);
```
kokokuo, line 82: Suggest to rename `LoadCacheAndSendActivityLog`

kokokuo (approved):
LGTM

PR #297, opened 2023-09-04
Feature: add RedShift

cyyeh:
## Description

Add Redshift as a data source

## Issue ticket number

None

PR #301, opened 2023-09-11
Feature: add redshift extension

cyyeh:
## Description

Basically, the PR content is the same as the closed one #297. However, I updated the content based on code review feedback from @kokokuo 

## Issue ticket number

None


packages/extension-driver-redshift/src/lib/redshiftDataSource.ts
```diff
@@ -0,0 +1,184 @@
...
+      }
+
+      const executeStatementCommand = new ExecuteStatementCommand(executeStatementCommandParams);
+      const statementCommandResult = await redshiftClient.send(executeStatementCommand);
+      return await this.getResultFromExecuteStatement(statementCommandResult, redshiftClient);
+    } catch (e: any) {
+      this.logger.debug(
+        `Errors occurred, release connection from ${profileName}`
+      );
+      redshiftClient.destroy();
```
oscar60310, line 92: Does `.destory` function here close the connection? We might reuse these clients so we should not close connections when error occurs.

packages/extension-driver-redshift/src/lib/redshiftDataSource.ts
```diff
@@ -0,0 +1,184 @@
...
+    redshiftClient: RedshiftDataClient
+  ): Promise<DataResult> {
+    let describeStatementResponse: DescribeStatementResponse | undefined;
+    const describeStatementRequestInput: DescribeStatementCommandInput = {
+      Id: statementCommandResult.Id,
+    };
+
+    // definition of describeStatementResponse.Status
+    // https://github.com/aws/aws-sdk-js-v3/blob/29056f4ca545f7e5cf951b915bb52178305fc305/clients/client-redshift-data/src/models/models_0.ts#L604
+    while (!describeStatementResponse || describeStatementResponse.Status !== 'FINISHED') {
```
oscar60310, line 133: Maybe we can add some exponential backoffs while polling status.

packages/extension-driver-redshift/src/lib/redshiftDataSource.ts
```diff
@@ -0,0 +1,184 @@
...
+      ) {
+        throw describeStatementResponse.Error
+      }
+    }
+
+    const getStatementResultCommandParams: GetStatementResultCommandInput = {
+      "Id": describeStatementResponse.Id
+    };
+    const getStatementResultCommand = new GetStatementResultCommand(getStatementResultCommandParams);
+    const getStatementResultResponse = await redshiftClient.send(getStatementResultCommand);
```
oscar60310, line 149: It looks like the result might return in batches, we ought to handle the `NextToken` of the response, store the token then fetch the next chunk data when we get request from downstream might be a good idea.

 ```ts
new Readable({
...
  read() {
    if(nextToken) { 
       // GetStatementResultCommand with nextToken
    }
  }
})

cyyeh: I've handled the `NextToken` condition at line 155.

packages/extension-driver-redshift/src/lib/redshiftDataSource.ts
cyyeh: I've deleted `.destroy` function

packages/extension-driver-redshift/src/lib/redshiftDataSource.ts
cyyeh: I added `exp-backoff`.

oscar60310: Thanks! I meant we can break between sending requests to avoid exceeding AWS API rate limiting. But I think your approach is okay, we'll retry automatically when the API returns an error.
```ts
  while (!describeStatementResponse || describeStatementResponse.Status !== 'FINISHED') {
      const describeStatementCommand = new DescribeStatementCommand(describeStatementRequestInput);
      describeStatementResponse = await redshiftClient.send(describeStatementCommand)

      if (
        describeStatementResponse.Status === 'ABORTED' || 
        describeStatementResponse.Status === 'FAILED'
      ) {
        throw describeStatementResponse.Error
      }

      // Sleep here before the next iteration
    }
```

packages/extension-driver-redshift/src/lib/redshiftDataSource.ts
oscar60310: Thanks, we won't lose data in this way.
If we encounter performance issues in the future, we could consider taking backpressure in Node.js streams into account~

oscar60310:
I have no further issues, many thanks!

kokokuo (approved):
Thanks @oscar60310 for reviewing!  ❤️ 
Thanks for fixing the test case issue, LGTM 👍

PR #310, opened 2023-09-18
Feature: extends CacheLayerInfo interface and Profile interface

onlyjackfrost:
## Description

Extends CacheLayerInfo interface and Profile interface to store properties that pass to the data source.

profile.yaml
```
 - name: duckdb
    type: duckdb
   connection:
      host: ...
      otherConnectionInfo
   allow: *
   // new added
   properties:
      some-configuration that can be read in the setting

```

API schema
```
urlPath: /artist/:id
cache:
 - cacheTableName: foo
   sql: 'select 1'
   profile: duckdb
   // new added
   options:
      foo: bar
profile: duckdb
```


kokokuo (approved):
LGTM 👍

PR #314, opened 2023-10-14
Fix: Fix API Parameter Type Bug Causing Undefined Error and Unintended Required Status

JSYOU:
## Description

This PR addresses a bug in the API definition file where specifying a parameter type would cause an error when the parameter was undefined, indirectly making it required. 
The fix ensures that specifying a type for a parameter no longer affects its required status.

## Issue ticket number
N/A
## Additional Context
``` yaml
urlPath: /artist/:id
request:
  - fieldName: id
    fieldIn: path
    description: constituent id
    validators:
      - required
  - fieldName: begin_date
    fieldIn: query
    type: number
    description: begin date
    validators:
      - date
sample:
  parameters:
    id: '1'
  profile: duckdb
profile: duckdb

```
before:
image

after:
image


kokokuo:
Hi @JSYOU,

Looks like you only solved one scenario for giving the type `number`, but when the user does not provide the `type` field, it happens when the schema.yaml created from Canner enterprise data service like the below source:

```yaml
# schema

urlPath: /artist/:id
request:
  - fieldName: id
    fieldIn: path
    description: constituent id
    validators:
      - required
  - fieldName: begin_date
    fieldIn: query
    description: begin date
    validators:
      - name: date
        args:
          format: YYYY-MM-DD
sample:
  # parameters:
  #   id: '1'
  # profile: duckdb
profile: duckdb
```

and the sql file

```sql
select 
*
from "artists"
where
ConstituentID = {{ context.params.id }}
{% if contex.params.begin_date %}
    and BeginDate = {{ context.params.begin_date }}
{% endif %}
```

PS: The code generated from the `labs/playground` package's artists but little changed schema.yaml

it still happens the error:

```
The input parameter is invalid, it should be date type
```

截圖 2023-10-14 下午11 58 49

kokokuo (approved):
LGTM 👍

kokokuo:
Only `date` validator needs to add the checker, and the `integer`, `uuid`, and `string` of the validator work, so it does not need to add the `undefined` checker.

PR #326, opened 2024-01-11
Fix: query again instead of using cached result when exporting from canner

### LINKED ISSUE #324: [Canner Datasource] Can choose to use cache or not when creating the caching data
## Environment

- Version:
  ```
  cli version: 10.0
  core version: 0.10.0
  build version: 10.0
  serve version: 10.0
  ```

- OS: macOS, windows 11 ...

## What's wrong?
I created an API using Vulcan with caching enabled, and the data source is Canner. However, I noticed that when Vulcan updates the cache, the data retrieved is not the latest. On examining the source code, I found that Vulcan uses temporary storage when updating and fetching data from Canner, resulting in non-up-to-date results.  

The result shoud be a up-to-date result

onlyjackfrost:
## Description

This PR fix the issue #324 according to the [Canner API Documentation](https://documenter.getpostman.com/view/21257531/2s93CGSc5k#8e91682d-5d3a-4f70-a2f5-9e90039a9c39) 

## Issue ticket number

closes #324

### DIFF
--- a/packages/extension-driver-canner/src/lib/cannerAdapter.ts
+++ b/packages/extension-driver-canner/src/lib/cannerAdapter.ts
@@ -44,6 +44,7 @@ export class CannerAdapter {
           sql,
           timeout: 600,
           noLimit: true,
+          cacheRefresh: true, // not use cached result
         },
       },
       headers


kokokuo (approved):
LGTM 👍

PR #332, opened 2024-01-18
Fix: fix incorrect condition to use restDataStream

### LINKED ISSUE #324: [Canner Datasource] Can choose to use cache or not when creating the caching data

onlyjackfrost:
## Description

The APIs with sql start with `WITH RECURSIVE` can not get response from server when the responding row is larger than the chunk size. 

## Issue ticket number

closes #324

### DIFF
--- a/packages/extension-api-caller/test/restApiCaller.spec.ts
+++ b/packages/extension-api-caller/test/restApiCaller.spec.ts
@@ -1,7 +1,6 @@
 import { getTestCompiler } from '@vulcan-sql/test-utility';
 import * as path from 'path';
 
-
 describe('Test "rest_api" filter', () => {
   it(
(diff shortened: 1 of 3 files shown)


kokokuo (approved):
LGTM 👍

PR #340, opened 2024-01-29
Feature: Model Drawer

andreashimin:
## Description

- Can create & update
- Need to confirm with Astro & William about custom field
- Need to confirm custom SQL behavior with William & Astro

image

image

image

## Issue ticket number

closes #xx


fredalai (approved):
LGTM

PR #343, opened 2024-01-30
Feature: mdl backend initialize

hanshino:
## Description

- migration scripts
- dummy api [(spec)](https://www.notion.so/canner/API-Spec-eb07b78eafde487ca6ccfa081b9122bd?pvs=4)
  - on boarding
  - modeling page (only support model CRUD)


onlyjackfrost (approved):
LGTM

PR #345, opened 2024-01-31
Feature: Metadata Drawer

andreashimin:
## Description

- View Metadata Drawer
- Update Metadata Popup
- The caculated field will remain update in model drawer which metadata drawer not provide update caculated fields, only can update it’s metadata.

image

image

image

## Issue ticket number

closes #xx

### DIFF
--- a/packages/admin-ui/components/table/MeasureFieldTable.tsx
+++ b/packages/admin-ui/components/table/MeasureFieldTable.tsx
@@ -1,0 +1,48 @@
+import { useMemo } from 'react';
+import { Table, TableProps } from 'antd';
+import { MeasureFieldValue } from '@vulcan-sql/admin-ui/components/modals/AddMeasureFieldModal';
+
+type Props = Pick<TableProps<MeasureFieldValue>, 'dataSource'> &
+  Partial<Pick<TableProps<MeasureFieldValue>, 'columns'>>;
+
+export const getMeasureFieldTableColumns = () => {
+  return [
+    {
+      title: 'Name',
+      dataIndex: 'fieldName',
+    },
+    {
+      title: 'Expression',
+      dataIndex: 'expression',
+      render: (expression, record) => {
+        // TODO: clarify the interface with backend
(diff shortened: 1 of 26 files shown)


packages/admin-ui/components/table/MeasureFieldTable.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { dataSource = [], columns = [] } = props;
+
+  const tableColumns = useMemo(
+    () => [...getMeasureFieldTableColumns(), ...columns],
+    [dataSource]
+  );
+
+  return (
+    <Table
+      rowKey={(record, index) => `${record.fieldName}-${index}`}
```
fredalai, line 37: For the current Ant.d V4 Table element, rowKey will no longer support **index**

andreashimin: I checked the latest `rc-table`, it supports `index`, though Antd doesn't write it out in their documents. 🤔️
https://github.com/react-component/table/blob/master/src/interface.ts#L118

image

fredalai: Hmm... I can see the `Warning: [antd: Table] `index` parameter of `rowKey` function is deprecated....` in the console.log. In fact, you will find that index is always 0.

fredalai: But this is not a big issue and does not affect the core functions. We can also improve it in subsequent PRs.

fredalai (approved):

packages/admin-ui/components/table/MeasureFieldTable.tsx
andreashimin: I saw the `Warning` after restarting my process 😢, still don't know why Antd validating it.
I use map to give the key property before table component. [a96a4e0](https://github.com/Canner/vulcan-sql/pull/345/commits/a96a4e07b02643fb3246557afcfc07549f308b68)

fredalai (approved):
LGTM

PR #349, opened 2024-02-07
Feature: Data modeling page

andreashimin:
## Description

- Integrating with all components
- Add Graphql dummy API
- Align naming with manifest API
- Adjust adapt data in modeling page
- Remove ui packages since there is no need to integrate with enterprise

image

image

image

image

image

image

## Issue ticket number

closes #xx


fredalai (approved):
LGTM

PR #358, opened 2024-02-27
Feature: add view drawer, metadata & refine modelling components

andreashimin:
## Description

- add view drawer
- add view metadata
- add view node
- create model drawer: remove fields select when in custom sql mode
- rename `caculated` -> `calculated`
- add code block component

## Issue ticket number

closes #xx


fredalai (approved):
LGTM

